PR #2, opened 2019-03-26
issue#1 fix

TomyYamy:
I tried commit #227df6e but it was not enough.
I find we need extra __init__.py in openai_gym and webots directory.


hadi88:
Yes, you're right! I have reported the same problem. We will fix this.

TomyYamy:
You changed the status "fix". However I think we had better to remain open it until it is fixed on main.

shromonag:
Thank you for including the __init__.py files. :)

PR #10, opened 2019-12-16
Upgrade webots to 2019

junjihashimoto:
This pr updates wbt files to support webots-2019.

* Update Superviser node to Robot node.
* Remove road textures webots-2019 does not have.


dfremont:
Closing this since we'll be updating the files directly to R2022a (in which the default coordinate system has changed, so the maps need to be re-imported). Apologies for never merging this!

PR #17, opened 2021-03-18
Branch 3 17

trevor-wu:
- Moved controller-related code to controller.py
- changed input of controller to (server, cache)


dfremont:
Closing for now -- we may revisit if code stabilizes.

PR #21, opened 2021-04-01
Branch 3 31

trevor-wu:
- Made it so controller returns only 
var_label_map = {'lat': lat, 'lon': lon, 'psis': psi, 'cte': cte, 'he' : heading_err}
- Append var_label_map to var_maps (new list)
- Generalized predicate function so that it can take in any var_label_map
- Refactored creation of simulation_data dictionary code
- Moved "stuck checker" to its own function so that no parameters are indexed in run_simulation function
@torfah


dfremont:
Closing with #17 for now.

PR #25, opened 2021-06-11
Kesav v/multi objective

kesav-v:
Multi-armed bandit sampling, parallelization, epsilon-greedy sampling, experiments framework


src/verifai/samplers/domain_sampler.py
```diff
@@ -25,6 +25,14 @@ class DomainSampler:
     def __init__(self, domain):
         self.domain = domain
 
+    def getSample(self):
+        """Generate the next sample, given the current distribution."""
+        return self.nextSample(feedback=None)
```
dfremont, line 30: Aren't `getSample` and `update` now the primary APIs, with `nextSample` only being used for backwards compatibility for sequential samplers? If so, I'd have `getSample` raise `NotImplementedError` and have `nextSample` be like this:
```
if self.last_sample is not None:
    assert feedback is not None
    self.update(self.last_sample, self.last_info, feedback)
self.last_sample, self.last_info = self.getSample()
return self.last_sample
```
This way, new samplers only need to implement `getSample` and `update`, but `nextSample` will still work if someone wants to use the old sequential API. Make sense?

src/verifai/samplers/domain_sampler.py
```diff
@@ -52,8 +60,27 @@ def __init__(self, domain, samplers):
         self.samplers = tuple(samplers)
 
     def nextSample(self, feedback=None):
-        return self.domain.rejoinPoints(
```
dfremont, line 55: Do we really need this? You should be able to implement `getSample` and `update` and then get `nextSample` for free, as indicated above.

src/verifai/samplers/domain_sampler.py
```diff
@@ -114,8 +141,13 @@ def __init__(self, domain):
         super().__init__(domain)
 
     def nextSample(self, feedback=None):
-        sample = self.nextVector(feedback)
-        return self.domain.unstandardize(sample)
+        sample, info = self.nextVector(feedback)
+        # print(f'inside BoxSampler.nextSample and sample = {sample}')
+        return self.domain.unstandardize(sample), info
```
dfremont, line 146: As above, this should not be needed.

src/verifai/samplers/domain_sampler.py
```diff
@@ -130,8 +162,12 @@ def __init__(self, domain):
         super().__init__(domain)
 
     def nextSample(self, feedback=None):
-        sample = self.nextVector(feedback)
-        return self.domain.unstandardize(sample)
+        sample, info = self.nextVector(feedback)
+        return self.domain.unstandardize(sample), info
```
dfremont, line 166: Likewise.

src/verifai/samplers/eg_sampler.py
```diff
@@ -0,0 +1,113 @@
...
+            elif isinstance(subsampler, DiscreteEpsilonGreedySampler):
+                assert self.disc_sampler is None
+                self.disc_sampler = subsampler
+            else:
+                assert isinstance(subsampler, RandomSampler)
+                assert self.rand_sampler is None
+                self.rand_sampler = subsampler
+
+    def nextSample(self, feedback=None):
+        return self.split_sampler.nextSample(feedback)
```
dfremont, line 46: Should not be needed.

src/verifai/samplers/eg_sampler.py
```diff
@@ -0,0 +1,113 @@
...
+
+    def nextVector(self, feedback=None):
+        self.update(None, self.current_sample, feedback)
+        return self.generateSample()
+    
+    def generateSample(self):
+        # proportions = self.errors / self.counts
+        # # print(proportions)
+        # Q = proportions + np.sqrt(2 / self.counts * np.log(1 + self.t*np.log(self.t)**2))
+        # bucket_samples = np.array([np.argmax(Q[i]) for i in range(len(self.buckets))])
```
dfremont, line 87: Remove unused code

src/verifai/samplers/domain_sampler.py
```diff
@@ -114,8 +141,13 @@ def __init__(self, domain):
...
-        return self.domain.unstandardize(sample)
+        sample, info = self.nextVector(feedback)
+        # print(f'inside BoxSampler.nextSample and sample = {sample}')
+        return self.domain.unstandardize(sample), info
+    
+    def getSample(self):
+        sample, info = self.nextVector()
+        return self.domain.unstandardize(sample), info
 
     def nextVector(self, feedback=None):
```
dfremont, line 152: Need to replace with `getVector` method and have `nextVector` be defined in terms of it, like `getSample` and `nextSample` in `DomainSampler`. Then new `BoxSampler` classes need only define `getVector` (and possibly `update` -- it probably makes sense to have an `updateVector` [or something] method too since `update` will receive the full sample, rather than the standardized vector which comes out of `getVector`, and this would be inconsistent).

src/verifai/samplers/eg_sampler.py
```diff
@@ -0,0 +1,113 @@
...
+        self.thres = thres
+        self.current_sample = None
+        self.counts = np.array([np.ones(int(b)) for b in buckets])
+        self.errors = np.array([np.zeros(int(b)) for b in buckets])
+        self.t = 1
+        self.epsilon = epsilon
+        self.sample_randomly = np.random.uniform() < self.epsilon
+
+    def nextVector(self, feedback=None):
+        self.update(None, self.current_sample, feedback)
```
dfremont, line 80: Should define `getVector` instead.

src/verifai/samplers/feature_sampler.py
```diff
@@ -61,6 +63,26 @@ def crossEntropySamplerFor(space, ce_params):
             lambda domain: CrossEntropySampler(domain=domain,
                                                ce_params=ce_params))
 
+    @staticmethod
+    def epsilonGreedySamplerFor(space, ce_params):
+        """Creates a cross-entropy sampler for a given space.
```
dfremont, line 68: Fix docstring.

src/verifai/samplers/feature_sampler.py
```diff
@@ -108,6 +130,14 @@ def makeDomainSampler(domain):
                 makeRandomSampler)
         return LateFeatureSampler(space, RandomSampler, makeDomainSampler)
 
+    def getSample(self):
+        """Generate the next sample, given the current distribution."""
+        return self.nextSample(feedback=None)
```
dfremont, line 135: This should be the primary API -- see comment on `DomainSampler.getSample`.

src/verifai/samplers/feature_sampler.py
```diff
@@ -163,15 +193,41 @@ def __init__(self, space, makeLengthSampler, makeDomainSampler):
 
     def nextSample(self, feedback=None):
         if self.lengthSampler is None:
```
dfremont, line 195: As above, this should not be necessary -- defining `getSample` and `update` should be enough.

src/verifai/samplers/feature_sampler.py
```diff
@@ -163,15 +193,41 @@ def __init__(self, space, makeLengthSampler, makeDomainSampler):
...
-            domainPoint = self.domainSamplers[length].nextSample(lastFeedback)
-        return self.space.makePoint(*domainPoint)
+            domainPoint, info2 = self.domainSamplers[length].nextSample(lastFeedback)
+            info = (info1, info2)
+        return self.space.makePoint(*domainPoint), info
+    
+    def getSample(self):
+        if self.lengthSampler is None:
+            domainPoint, info = self.domainSampler.getSample()
+            # print(f'domainPoint = {domainPoint}')
```
dfremont, line 211: Remove debugging print

src/verifai/samplers/multi_armed_bandit.py
```diff
@@ -0,0 +1,203 @@
...
+                self.cont_sampler = subsampler
+            elif isinstance(subsampler, DiscreteMultiArmedBanditSampler):
+                assert self.disc_sampler is None
+                self.disc_sampler = subsampler
+            else:
+                assert isinstance(subsampler, RandomSampler)
+                assert self.rand_sampler is None
+                self.rand_sampler = subsampler
+
+    def nextSample(self, feedback=None):
```
dfremont, line 48: Should not be necessary

src/verifai/samplers/multi_armed_bandit.py
```diff
@@ -0,0 +1,203 @@
...
+        self.errors = np.array([np.zeros(int(b)) for b in buckets])
+        self.t = 1
+        self.counterexamples = dict()
+        self.is_multi = False
+        self.invalid = np.array([np.zeros(int(b)) for b in buckets])
+        self.monitor = None
+        self.rho_values = []
+        self.restart_every = restart_every
+
+    def nextVector(self, feedback=None):
```
dfremont, line 86: Replace with `getVector`

src/verifai/samplers/multi_armed_bandit.py
```diff
@@ -0,0 +1,203 @@
...
+        # choose the bucket with the highest "goodness" value, breaking ties randomly.
+        bucket_samples = np.array([np.random.choice(np.flatnonzero(np.isclose(Q[i], Q[i].max())))
+            for i in range(len(self.buckets))])
+        self.current_sample = bucket_samples
+        ret = tuple(np.random.uniform(bs, bs+1.)/b for b, bs
+              in zip(self.buckets, bucket_samples))
+        return ret, bucket_samples
+    
+    def update(self, sample, info, rho):
+        if rho is None:
```
dfremont, line 102: This should not be possible. `update` should only be called after a sample has been evaluated.

src/verifai/samplers/multi_armed_bandit.py
```diff
@@ -0,0 +1,203 @@
...
+            self.counts[i][b] += 1.
+            if rho < self.thres:
+                self.errors[i][b] += 1.
+        # print(self.errors / self.counts)
+
+    def set_graph(self, graph):
+        self.priority_graph = graph
+        try:
+            self.thres = [self.thres] * graph.number_of_nodes()
+        except Exception as e:
```
dfremont, line 123: This looks odd -- see corresponding comment in multi-CE sampler.

src/verifai/samplers/multi_armed_bandit.py
```diff
@@ -0,0 +1,203 @@
...
+                to_remove.add(other_ce)
+        for other_ce in to_remove:
+            del self.counterexamples[other_ce]
+        self.counterexamples[ce] = np.array([np.zeros(int(b)) for b in self.buckets])
+        return True
+    
+    def update_dist_from_multi(self, sample, info, rho):
+        try:
+            iter(rho)
+        except:
```
dfremont, line 180: Should catch a specific `Exception` here (or use `isinstance` instead) -- currently this will catch everything, including `KeyboardInterrupt`!

experiments/README.md
```diff
@@ -0,0 +1,15 @@
+# Installation of VerifAI and Scenic
+
+1. Clone this repository and [Scenic](https://github.com/BerkeleyLearnVerify/Scenic).
+2. Checkout the `kesav-v/multi-objective` branch of VerifAI and the `kesav-v/multi-objective` branch of Scenic.
```
dfremont, line 4: Update this (and tricky bit below about doing the installs in the right order) assuming we have a new release of Scenic which integrates your code. Then just cloning VerifAI and installing as usual should be enough.

experiments/README.md
```diff
@@ -0,0 +1,15 @@
...
+3. Install `poetry` if you haven’t already done so.
+4. Run `poetry shell` from the VerifAI repo and make sure it spawns an environment with Python 3.8+.
+5. Run `poetry install`.
+6. Go to the location where `Scenic` was cloned and run `poetry install` (while in the same environment that was used for VerifAI).
+7. Any other missing packages when running the falsifier script can be installed using `pip`.
```
dfremont, line 9: Any dependencies required for the VerifAI examples should be added to pyproject.toml as optional dependencies and listed under the "examples" extra. Then replacing the above command with `poetry install -E examples` will install all needed dependencies.

experiments/README.md
```diff
@@ -0,0 +1,15 @@
...
+
+# Running the Newtonian Simulator
+
+1. Using the VerifAI environment, go to the `experiments` folder inside VerifAI
```
dfremont, line 13: As we discussed, let's move this under the examples folder.

src/verifai/samplers/multi_armed_bandit.py
```diff
@@ -0,0 +1,203 @@
...
+            for i, b in enumerate(info):
+                self.invalid[i][b] += 1.
+            return
+        # print('inside update_dist_from_multi')
+        counter_ex = tuple(
+            rho[node] < self.thres[node] for node in nx.dfs_preorder_nodes(self.priority_graph)
+        )
+        self.rho_values.append(counter_ex)
+        # print(f'counter_ex = {counter_ex}')
+        # print(self.counterexamples)
```
dfremont, line 194: Remove debugging code

src/verifai/samplers/multi_objective.py
```diff
@@ -0,0 +1,18 @@
+from abc import ABC, abstractmethod
+
+class MultiObjectiveSampler:
+
+    def set_priority_graph(self, graph):
+        self.priority_graph = graph
+
+    """
+    Update distribution from list of objectives.
```
dfremont, line 9: Docstring should be inside function definition. Also should not assume you're using `ScenicSampler`.

src/verifai/samplers/multi_objective.py
```diff
@@ -0,0 +1,18 @@
...
+
+    Arguments:
+        sample: the sample as outputted by ScenicSampler.nextSample()
+        info: Contains any necessary information to perform the update (such as bucket indices, etc.)
+        rho: A NumPy array with the real objective values [f1, f2, f3, ..., fn]
+    """
+    @abstractmethod
+    def update_dist_from_multi(self, sample, info, rho):
```
dfremont, line 17: Why do we need this function? Why not just use the ordinary `update` and have `rho` be a tuple of values?

src/verifai/samplers/scenic_sampler.py
```diff
@@ -104,12 +104,14 @@ def pointForValue(dom, scenicValue):
 ignoredParameters = {
     'externalSampler', 'externalSamplerRejectionFeedback',
     'verifaiSamplerType', 'verifaiSamplerParams',
+    'behavior',
```
dfremont, line 107: Don't need this -- this says to ignore a global parameter called "behavior".

src/verifai/samplers/scenic_sampler.py
```diff
@@ -210,9 +212,9 @@ def __init__(self, scenario, maxIterations=None, ignoredProperties=None):
         super().__init__(space)
 
     @classmethod
-    def fromScenario(cls, path, maxIterations=None, ignoredProperties=None):
+    def fromScenario(cls, path, maxIterations=None, ignoredProperties=None, model=None, **params):
```
dfremont, line 215: I would not have `params` be a set of keyword arguments here: we should separate arguments that `fromScenario` needs from global parameters that should be passed through to the scenario.

src/verifai/scenic_server.py
```diff
@@ -23,25 +30,29 @@ def __init__(self, sampling_data, monitor, options={}):
         else:
             self.rejectionFeedback = extSampler.rejectionFeedback
         self.monitor = monitor
+        if isinstance(self.monitor, multi_objective_monitor):
+            self.sampler.scenario.externalSampler.sampler.domainSampler.split_sampler.samplers[0].set_graph(self.monitor.graph)
```
dfremont, line 34: This assumes the Scenic program is using an external sampler, and one of a particular form. Need a more generic solution to this. I think when we discussed this earlier I suggested passing the graph to the `ScenicSampler` and having the samplers pass this information down recursively (ideally when the samplers are being initialized, rather than calling something like `set_graph` after the fact).

src/verifai/scenic_server.py
```diff
@@ -23,25 +30,29 @@ def __init__(self, sampling_data, monitor, options={}):
...
         assert scene
+        after_sampling = time.time()
         result = self._simulate(scene)
         if result is None:
             self.lastValue = self.rejectionFeedback
         else:
             self.lastValue = (0 if self.monitor is None
-                              else self.monitor.evaluate(result.trajectory))
-        return sample, self.lastValue
+                              else self.monitor.evaluate(result))
```
dfremont, line 53: Why pass the entire `Simulation` to the monitor? Monitors expect a trajectory - has that changed?

src/verifai/scenic_server.py
```diff
@@ -62,3 +73,160 @@ def _simulate(self, scene):
...
+
+    def nextSample(self, feedback):
+        return self.last_sample, None
+
+@ray.remote
+class SampleSimulator():
+
+    def __init__(self, scenic_path, worker_num, monitor, options={}, use_carla=False,
+    scenario_params={}):
+        print(scenario_params)
```
dfremont, line 87: Remove debugging print

src/verifai/scenic_server.py
```diff
@@ -62,3 +73,160 @@ def _simulate(self, scene):
...
+        scenario_params.update({
+            'port': 2000 + 2*worker_num
+        })
```
dfremont, line 90: Overriding a global parameter named "port" with this particular value needs to be documented.

src/verifai/scenic_server.py
```diff
@@ -62,3 +73,160 @@ def _simulate(self, scene):
...
+        self.worker_num = worker_num
+        self.sampler.scenario.externalSampler = DummySampler(self.sampler.scenario.externalParams,
+        self.sampler.scenario.params)
+        self.simulator = self.sampler.scenario.getSimulator()
+        self.monitor = monitor
+        extSampler = self.sampler.scenario.externalSampler
+        if extSampler is None:
+            self.rejectionFeedback = None
+        else:
+            self.rejectionFeedback = extSampler.rejectionFeedback
```
dfremont, line 103: Do we need any of this rejection stuff here, since you're using the DummySampler?

src/verifai/scenic_server.py
```diff
@@ -62,3 +73,160 @@ def _simulate(self, scene):
...
+        except SimulationCreationError as e:
+            if self.verbosity >= 1:
+                print(f'  Failed to create simulation: {e}')
+            self.lastValue = self.rejectionFeedback
+            return self.worker_num, self.full_sample, self.lastValue
+        except RuntimeError as e:
+            print(f'Runtime error during simulation: {e}')
+            print('Waiting 1 minute before continuing...')
+            time.sleep(60)
+            return self.worker_num, self.full_sample, self.lastValue
```
dfremont, line 139: What is this code supposed to handle? Catching all RuntimeErrors is pretty broad.

src/verifai/scenic_server.py
```diff
@@ -62,3 +73,160 @@ def _simulate(self, scene):
...
+        print(f'Sampler class is {type(self.sampler)}')
+        self.sample_simulators = [SampleSimulator.remote(scenic_path, i, monitor, options,
+        use_carla, scenario_params)
+        for i in range(self.total_workers)]
+
+    def _generate_next_sample(self, worker_num):
+        i = 0
+        feedback = self.lastValue
+        ext = self.sampler.scenario.externalSampler
+        while i < 2000:
```
dfremont, line 171: What is this hard-coded limit?

src/verifai/scenic_server.py
```diff
@@ -62,3 +73,160 @@ def _simulate(self, scene):
...
+        sampling_data.sampler = sampler
+        super().__init__(sampling_data, monitor, options)
+        print(f'Sampler class is {type(self.sampler)}')
+        self.sample_simulators = [SampleSimulator.remote(scenic_path, i, monitor, options,
+        use_carla, scenario_params)
+        for i in range(self.total_workers)]
+
+    def _generate_next_sample(self, worker_num):
+        i = 0
+        feedback = self.lastValue
```
dfremont, line 169: Doesn't look like this is getting used any more (I guess the update is happening elsewhere?)

src/verifai/scenic_server.py
```diff
@@ -62,3 +73,160 @@ def _simulate(self, scene):
...
+            futures.append(sim.simulate.remote(next_sample))
+        while True:
+            done, _ = ray.wait(futures)
+            result = ray.get(done[0])
+            t = time.time() - startTime
+            index, sample, rho = result
+            self.lastValue = rho
+            results.append((sample, rho))
+            info = infos[index]
+            self.sampler.scenario.externalSampler.update(sample, info, rho)
```
dfremont, line 216: What if the program doesn't use an external sampler?

src/verifai/server.py
```diff
@@ -20,6 +21,7 @@ def default_sampler_params(sampler_type):
 
 def choose_sampler(sample_space, sampler_type,
                    sampler_params=None):
+    print(f'sampler type is {sampler_type}')
```
dfremont, line 24: Remove debugging prints here and below

src/verifai/scenic_server.py
```diff
@@ -62,3 +73,160 @@ def _simulate(self, scene):
...
+
+class DummySampler(VerifaiSampler):
+
+    def nextSample(self, feedback):
+        return self.last_sample, None
+
+@ray.remote
+class SampleSimulator():
+
+    def __init__(self, scenic_path, worker_num, monitor, options={}, use_carla=False,
```
dfremont, line 85: The `use_carla` argument seems to be unused.

tests/test_scenic.py
```diff
@@ -5,7 +5,7 @@
 from verifai.samplers.scenic_sampler import ScenicSampler
 from verifai.scenic_server import ScenicServer
 from verifai.falsifier import generic_falsifier
-from tests.utils import sampleWithFeedback, checkSaveRestore
+# from utils import sampleWithFeedback, checkSaveRestore
```
dfremont, line 8: Did you run these tests? I think commenting out this import will break them...

tests/test_scenic.py
```diff
@@ -129,3 +129,4 @@ def test_driving_dynamic(pathToLocalFile):
                                   server_class=ScenicServer,
                                   server_options=server_options)
     falsifier.run_falsifier()
+    print('end of test')
```
dfremont, line 132: Any need for this print? `pytest` will suppress it by default.

src/verifai/falsifier.py
```diff
@@ -1,14 +1,26 @@
...
+def parallelized(server_class):
+    if server_class == Server:
+        return ParallelServer
+    elif server_class == ScenicServer:
+        return ParallelScenicServer
 
 class falsifier(ABC):
     def __init__(self, monitor, sampler_type=None, sampler=None, sample_space=None,
-                 falsifier_params=None, server_options={}, server_class=Server):
+                 falsifier_params=None, server_options={}, server_class=Server, use_carla=False):
```
dfremont, line 23: The argument `use_carla` appears to be unused -- in any case obviously the generic falsifier class should not have any simulator-specific arguments (communication with the simulator is the job of the Server).

dfremont (changes requested):
Looks good overall -- see comments below for various things to fix, mostly minor (let me know if anything is unclear). Please also delete "poetry.lock". Thanks!

pyproject.toml
```diff
@@ -39,7 +39,6 @@ pandas = "^1.0.3"
 scikit-learn = "^0.22.2"
 pygame = "^2.0.0.dev6"
 kmodes = "^0.10.2"
-scenic = "^2.0.0b1"
```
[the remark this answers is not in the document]
kesav-v: Would that work? I believe there are some changes I made to Scenic that are necessary for this branch of VerifAI to work. Are those already merged in and being used in the latest PyPI version of Scenic?

run_scenic.py
```diff
@@ -0,0 +1,56 @@
+
```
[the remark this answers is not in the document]
kesav-v: @ek65 can you take a look at this when you get a chance? I checked `git blame` and it looks like you wrote this file, so if you could let me know where to move this that'd be great.

src/verifai/falsifier.py
```diff
@@ -98,40 +116,73 @@ def analyze_error_table(self, analysis_params= None, error=None):
...
+            i += 1
+            bar.update(i)
+            if i == 1:
+                t0 = time.time()
+            if self.n_iters is not None and i == self.n_iters:
+                break
+            if self.max_time is not None and time.time() - t0 >= self.max_time:
+                break
+        if isinstance(self.monitor, multi_objective_monitor):
+            counterexamples = self.server.sampler.scenario.externalSampler.sampler.domainSampler.split_sampler.samplers[0].counterexample_values
```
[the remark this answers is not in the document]
kesav-v: It should work in general as well - it only overwrites the existing value of `counterexamples` if multi-dimensional objectives are used, and that is only supported by the multi-armed bandit sampler for now. So if any other sampler is used, `self.monitor` will not be an instance of `multi_objective_monitor` and this line will never run.

src/verifai/falsifier.py
```diff
@@ -98,40 +116,73 @@ def analyze_error_table(self, analysis_params= None, error=None):
...
                 if self.verbosity >= 1:
                     print("Sampler has generated all possible samples")
                 break
-            if self.verbosity >= 1:
-                print("Sample no: ", i, "\nSample: ", sample, "\nRho: ", rho)
+            # if self.verbosity >= 1:
+            #     print("Sample no: ", i, "\nSample: ", sample, "\nRho: ", rho)
             self.samples[i] = sample
-            if isinstance(rho, (list, tuple)):
-                check_var = rho[-1]
```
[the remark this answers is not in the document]
kesav-v: Yes - whether or not the rho value corresponds to a counterexample needs to be directly handled by the sampler now, since rho being a list or tuple could also mean that the monitor is a multi_objective_monitor.

PR #55, opened 2025-02-13
Update dependency constraints for Python 3.13 support

lola831:
This PR updates dependency constraints to ensure compatibility with Python 3.13.

-**Pandas:** Updated to allow version 2.x.
-**Pillow:** Adjusted to support newer releases compatible with Python 3.13.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,7 +26,7 @@ classifiers = [
 ]
 
 dependencies = [
-	"numpy ~= 1.21",
+	"numpy >= 1.21",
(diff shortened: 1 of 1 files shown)


dfremont (changes requested):
Looks good except we should allow `numpy >= 1.21` (rather than `~=`).

lola831:
Made the suggested change to numpy >= 1.21 (rather than ~=).

dfremont (approved):
Thanks!

PR #57, opened 2025-11-11
Add compositional analysis code and example

beyazit-y:


kevinchang73 (approved):

kevinchang73 (approved):

examples/compositional_analysis/README.md
```diff
@@ -0,0 +1,14 @@
...
+
+The core idea is to perform **Statistical Model Checking (SMC)** or **falsification** on *primitive scenarios* — scenarios that serve as building blocks for defining more complex *composite scenarios*. The analysis traces generated from these primitives are stored in a `ScenarioBase` object.
+
+These traces can then be supplied to an instance of `CompositionalAnalysisEngine`, which supports querying over composite scenario structures to perform compositional analysis based on the primitive scenario traces.
+
+This example uses the [MetaDrive](https://metadriverse.github.io/metadrive/) simulator.
+
+* Train a reinforcement learning policy using `train.py`.
+* Test the policy and save generated traces using `test.py`.
+* See `analyze.py` for an example of how to perform compositional analysis on generated traces.
```
dfremont, line 13: If I understand correctly, just running these three scripts in a row won't work: `analyze.py` looks for files which are only created by `run_exp.py`. Please give an example of how to run that script. (Or do you want people to run `exps.sh` instead? That file isn't explained either.)

examples/compositional_analysis/run_exp.py
```diff
@@ -0,0 +1,357 @@
...
+    Args:
+        confidence_level: Desired confidence level (e.g., 0.95 for 95%)
+        error_bound: Maximum error epsilon (e.g., 0.01 for 1%)
+    
+    Returns:
+        Number of samples needed
+    """
+    delta = 1 - confidence_level
+    n = math.log(2 / delta) / (2 * error_bound ** 2)
+    return int(math.ceil(n))
```
dfremont, line 28: `math.ceil` already returns an `int`

examples/compositional_analysis/run_exp.py
```diff
@@ -0,0 +1,357 @@
...
+    for s in scenarios:
+        scenario_dir = os.path.join(save_dir, s)
+        if os.path.exists(scenario_dir):
+            shutil.rmtree(scenario_dir)
+    
+    print("=== Generating Traces (Parallel - HARD STOP) ===")
+    
+    # Launch all processes
+    processes = []
+    start_time = time.time()
```
dfremont, line 64: Timings should use `time.monotonic()` so as not to be affected by changes to the system clock.

examples/compositional_analysis/run_exp.py
```diff
@@ -0,0 +1,357 @@
...
+                if len(lines) <= 1:  # Only header or empty
+                    current_count = 0
+                else:
+                    # Count unique trace_ids in final file
+                    trace_ids = set()
+                    for line in lines[1:]:  # Skip header
+                        parts = line.split(',')
+                        if parts:
+                            trace_ids.add(parts[0])
+                    current_count = len(trace_ids)
```
dfremont, line 145: This code looks basically identical to the code in the previous loop -- please consolidate into a helper function.

examples/compositional_analysis/run_exp.py
```diff
@@ -0,0 +1,357 @@
...
+                # Only add to logs if there are any complete traces
+                if expected_count > 0:
+                    logs[s] = csv_path
+                    print(f"[INFO] Scenario {s} has {expected_count} completed episodes.")
+                else:
+                    print(f"[INFO] Scenario {s} had no completed episodes.")
+            else:
+                # Process completed successfully
+                # Count episodes in completed file
+                with open(csv_path, 'r') as f:
```
dfremont, line 172: Likewise, let's use a helper function (maybe the same one) for all this CSV reading stuff.

examples/compositional_analysis/run_exp.py
```diff
@@ -0,0 +1,357 @@
...
+        )
+
+        print(f"Estimated {s}: rho = {rho:.4f} ± {uncertainty:.4f}")
+
+        results[s] = {"rho": rho, "uncertainty": uncertainty}
+
+    return results
+
+
+def parse_scenario(input_scenario):
```
dfremont, line 241: What is the purpose of this function? It looks equivalent to `set(input_scenario)`.

examples/compositional_analysis/run_exp.py
```diff
@@ -0,0 +1,357 @@
...
+        scenario_base = ScenarioBase(logs, delta=1-confidence_level)
+        
+        # compositional rho
+        run_SMC_compositional(scenarios=[input_scenario], time_budget=time_budget, scenario_base=scenario_base)
+    
+    # Print summary for easy copy-paste to README
+    print("\n" + "="*60)
+    print("SUMMARY")
+    print("="*60)
+    print(f"Command: python compare_analysis.py --scenario \"{input_scenario}\" "
```
dfremont, line 308: This would be easier to read as a triple-quoted f-string (you wouldn't need to escape the quotes or keep writing `f"` at the start of each line).

examples/compositional_analysis/run_exp.py
```diff
@@ -0,0 +1,357 @@
...
+          f"{'--compositional ' if isCompositional else ''}"
+          f"{'--expert ' if expert else ''}"
+          f"{'--reuse_traces ' if reuse_traces else ''}"
+          f"--time_budget {time_budget if time_budget != float('inf') else 'N/A'} "
+          f"--save_dir \"{save_dir}\"")
+    print("="*60)
+
+
+if __name__ == "__main__":
+    import argparse
```
dfremont, line 318: Should be at top of file (there's no need to defer this import).

examples/compositional_analysis/run_exp.py
```diff
@@ -0,0 +1,357 @@
...
+    parser.add_argument("--confidence_level", type=float, default=None, help="Confidence level for ground truth (default: None)")
+    parser.add_argument("--error_bound", type=float, default=None, help="Error bound (epsilon) for ground truth (default: None)")
+    parser.add_argument("--reuse_traces", action="store_true", help="Use existing traces from save_dir without generating new ones (default: False)")
+    
+    args = parser.parse_args()
+    
+    mp.set_start_method("spawn")
+
+    assert args.confidence_level is not None
+    assert (args.error_bound is not None) or (args.time_budget is not None)
```
dfremont, line 336: These are provided by the user, so you should raise an exception if they are invalid, not fail an assertion. For `confidence_level`, if it is required you can just tell `argparse` that.

examples/compositional_analysis/test.py
```diff
@@ -0,0 +1,167 @@
...
+warnings.filterwarnings("ignore", message="pkg_resources is deprecated")
+
+import os
+import csv
+import argparse
+import numpy as np
+from train import make_env
+from functools import partial
+import matplotlib.pyplot as plt
+from stable_baselines3 import PPO
```
dfremont, line 11: New dependencies such as this one will need to be added to the `examples` extra in `pyproject.toml` so that `pip install verifai[examples]` gets everything needed to run the examples.

examples/compositional_analysis/test.py
```diff
@@ -0,0 +1,167 @@
...
+from metadrive.envs import MetaDriveEnv
+from IPython.display import Image, clear_output
+from metadrive.utils.doc_utils import generate_gif
+from metadrive.component.map.base_map import BaseMap
+from metadrive.policy.expert_policy import ExpertPolicy
+from stable_baselines3.common.utils import set_random_seed
+from metadrive.component.map.pg_map import MapGenerateMethod
+from metadrive.utils.draw_top_down_map import draw_top_down_map
+from stable_baselines3.common.vec_env.subproc_vec_env import SubprocVecEnv
```
dfremont, line 20: In this and all other files, please arrange imports in 3 blocks separated by 1 blank line: Python stdlib packages, external packages, and imports from VerifAI itself.

examples/compositional_analysis/test.py
```diff
@@ -0,0 +1,167 @@
...
+    #     # ret = env.render(mode="topdown",
+    #     #                  window=False,
+    #     #                  # screen_size=(600, 600),
+    #     #                  # camera_position=(50, 50)
+    #     #                  )
+    #     env.close()
+    #     plt.axis("off")
+    #     plt.imshow(ret)
+    #     plt.show()
+    #     clear_output()
```
dfremont, line 70: Either delete this or move it into a separate function for debugging.

examples/compositional_analysis/test.py
```diff
@@ -0,0 +1,167 @@
...
+
+
+    all_traces = []
+    trace_id = 0
+
+    if not args.gif:
+        csv_path = os.path.join(args.save_dir, "traces.csv")
+        if os.path.exists(csv_path):
+            os.remove(csv_path)
+        f = open(csv_path, "w", newline="")
```
dfremont, line 85: Should use `finally:` to ensure this file is closed even if an exception is raised. You'll probably want to move the main loop into a helper function so that the `try:` isn't huge.

examples/compositional_analysis/train.py
```diff
@@ -0,0 +1,107 @@
...
+        default=1_000_000,
+        help="Number of environment steps")
+    parser.add_argument(
+        "--scenario",
+        type=str,
+        default="2",
+        help="Scenario string")
+    args = parser.parse_args()
+
+    # while True:
```
dfremont, line 74: Remove debugging code

examples/compositional_analysis/utils.py
```diff
@@ -0,0 +1,130 @@
...
+        save_dir (str): Directory where traces or gifs will be saved.
+        model_path (str): Path to the trained PPO model (.zip file). Not used if expert=True.
+        expert (bool): If True, use expert policy instead of trained model.
+        n (int): Number of test episodes to run.
+        scenario (str or int): Scenario string or ID.
+        gif (bool): If True, generate top-down gifs instead of CSV traces.
+    """
+
+    if not expert:
+        assert model_path is not None, "You must provide a valid model_path (.zip file)"
```
dfremont, line 32: Raise `ValueError` instead

examples/compositional_analysis/utils.py
```diff
@@ -0,0 +1,130 @@
...
+
+    set_random_seed(seed)
+
+    scenario_id = int(scenario) if str(scenario).isdigit() else scenario
+    env = make_env(scenario=scenario_id, monitor=False)
+    
+    if expert:
+        # print("USING EXPERT POLICY")
+        from metadrive.policy.expert_policy import ExpertPolicy
```
dfremont, line 41: Is there a reason to defer this import? If not, it should be at the top of the file.

examples/compositional_analysis/utils.py
```diff
@@ -0,0 +1,130 @@
...
+            env.top_down_renderer.generate_gif(gif_path)
+            print(f"Saved gif to {gif_path}")
+
+        trace_id += 1
+
+    if not gif:
+        f.close()
+
+    env.close()
+    return
```
dfremont, line 129: Redundant

examples/compositional_analysis/utils.py
```diff
@@ -0,0 +1,130 @@
+import os
+import csv
+import numpy as np
+from stable_baselines3 import PPO
+from stable_baselines3.common.utils import set_random_seed
+from train import make_env
+
+
+def generate_traces(
```
dfremont, line 9: This function seems almost the same as the main code in `test.py`. Can you use it there, with appropriate options? If not, it would be good to consolidate as much as possible into common helper functions.

src/verifai/compositional_analysis.py
```diff
@@ -0,0 +1,331 @@
...
+            df_s, df_t = self.scenario_base.data[s_name], self.scenario_base.data[t_name]
+
+            # Select successful endpoints
+            s_last = df_s.sort_values("step").groupby("trace_id").tail(1)
+            s_last = s_last[s_last["label"] == True]
+            t_first = df_t.sort_values("step").groupby("trace_id").head(1)
+            t_last = df_t.sort_values("step").groupby("trace_id").tail(1)
+
+            # KDE features
+            if features:
```
dfremont, line 126: Promote this check out of the loop (i.e. write `if not features:` and raise the exception before the loop).

src/verifai/compositional_analysis.py
```diff
@@ -0,0 +1,331 @@
...
+
+                fail_idx = (t_last["label"] == False).to_numpy()
+                t_first = t_first[fail_idx].sort_values("trace_id")
+                t_last = t_last[fail_idx].sort_values("trace_id")
+
+                if t_first.empty or t_last.empty:
+                    continue
+
+            # KDE features
+            if features:
```
dfremont, line 240: As above. In fact, can some of the code of `check` and `falsify` be unified?

examples/compositional_analysis/train.py
```diff
@@ -0,0 +1,107 @@
...
+    parser.add_argument(
+        "--model",
+        type=str,
+        default=None,
+        help="Model zip name"
+    )
+    parser.add_argument(
+        "--n-envs",
+        type=int,
+        default=16,
```
dfremont, line 60: This is a big default (remember people may try this on their laptops). I'd suggest either a smaller constant, like 4, or using the total number of logical cores on the system minus 1.

examples/compositional_analysis/README.md
```diff
@@ -0,0 +1,14 @@
...
+This directory provides an example use case of **compositional analysis** for *Markovian* (i.e., memoryless) specifications.
+
+The core idea is to perform **Statistical Model Checking (SMC)** or **falsification** on *primitive scenarios* — scenarios that serve as building blocks for defining more complex *composite scenarios*. The analysis traces generated from these primitives are stored in a `ScenarioBase` object.
+
+These traces can then be supplied to an instance of `CompositionalAnalysisEngine`, which supports querying over composite scenario structures to perform compositional analysis based on the primitive scenario traces.
+
+This example uses the [MetaDrive](https://metadriverse.github.io/metadrive/) simulator.
+
+* Train a reinforcement learning policy using `train.py`.
+* Test the policy and save generated traces using `test.py`.
```
dfremont, line 12: Running this I got the following warning which would be good to investigate:
```
[WARNING] env.vehicle will be deprecated soon. Use env.agent instead (base_env.py:734)
```

src/verifai/compositional_analysis.py
```diff
@@ -0,0 +1,331 @@
...
+        std = np.std(features, axis=0)
+        std[std == 0] = 1.0  # Avoid division by zero
+        return (features - mean) / std
+
+    def check(
+        self,
+        scenario: List[str],
+        features: Optional[List[str]] = None,
+        center_feat_idx: Optional[List[int]] = None,
+        bw_method: str | int = 10,
```
dfremont, line 86: This type union syntax was not added until Python 3.10, but VerifAI supports 3.8 and 3.9 (and in fact Alejandro's example requires 3.9). Please fix and test your code on 3.8 and 3.9. (This would have been caught by `tox` if you had any test cases, which I'll mention in the top-level review.)

examples/compositional_analysis/README.md
dfremont: Even after running `exps.sh` I get the following error from `analyze.py`:
```
Traceback (most recent call last; use -b to show Scenic internals):
  File "/Users/dfremont/Documents/Berkeley/repositories/BerkeleyLearnVerify/VerifAI/examples/compositional_analysis/analyze.py", line 18, in <module>
    scenario_base = ScenarioBase(logs, delta=0.01)
  File "/Users/dfremont/Documents/Berkeley/repositories/BerkeleyLearnVerify/VerifAI/src/verifai/compositional_analysis.py", line 38, in __init__
    raise FileNotFoundError(f"CSV file for scenario '{name}' not found: {path}")
FileNotFoundError: CSV file for scenario 'S' not found: storage/traces/S/traces.csv
```

dfremont (changes requested):
Hi Beyazit, thanks for making this PR! With a few changes I was able to run everything except the final analysis (see comments), and the code looks OK overall. A few bigger things to improve:

* There should be unit tests for the new functionality. They can be basic, not using a neural network or anything (we want the tests to run without optional dependencies such as stable baselines anyway), but should exercise the main code paths and check that the output is reasonable.
* There should be a new page in the documentation describing the compositional analysis features. It can explain the example in more detail (for example, what do `S`, `X`, `SXS`, etc. actually mean) and also explain how to use the top-level APIs `ScenarioBase` and `CompositionalAnalysisEngine`. The docstrings for those are a good starting point (and you can auto-generate documentation from them, as we do in `docs/feature_api.rst` for example), but they need to be fleshed out: for example, the analysis methods accept a list of scenario names, but the example only ever calls them with a single string.

If you can help me solve the issue I had when running `analyze.py` I will try to run the example again. Thanks!

examples/compositional_analysis/train.py
```diff
@@ -0,0 +1,107 @@
...
+    parser = argparse.ArgumentParser(description="Train policy in MetaDrive")
+    parser.add_argument(
+        "--seed",
+        type=int,
+        default=0,
+        help="Random seed for reproducibility")
+    parser.add_argument(
+        "--save-dir",
+        type=str,
+        default="storage",
```
dfremont, line 49: Let's also add a `.gitignore` file in this folder saying to ignore `storage`; otherwise running the example makes Git think you've made a bunch of changes.

examples/compositional_analysis/README.md
beyazit-y: I fixed the scripts and README.md. Also, included a small set of traces to make example.py runnable.

examples/compositional_analysis/run_exp.py
beyazit-y: We included this as a place because, in the long term, the goal is to parse Scenic files. But it is redundant for this file, so I'll remove it.

beyazit-y:
I fixed most of the comments. Specifically: (i) fixed the example (it should run smoothly now -- lmk if it doesn't work), (ii) removed redundant code, (iii) eliminated most of the repetitive code, (iv) added unit tests, and (v) improved README and docstrings. There is still more we can improve esp `check` and `falsify` methods, which can be further reimplemented to eliminate code repetition; however, as the `falsify` method is still experimental (which is mentioned in the README), I think it is good to keep those separate for now. I am working with an undergraduate to improve the code in this regard. Please lmk if you have other comments. Thanks!

dfremont (approved):
Thanks, Beyazit, it's looking good. I'm going to merge this into a branch where I'll be putting all the other PRs, so that I can make sure everything works well together before merging into `main`.

PR #58, opened 2025-11-13
VerifAI Dynamic Sampling

Eric-Vin:
Adds dynamic sampling to VerifAI.

Merge with https://github.com/BerkeleyLearnVerify/Scenic/pull/417

### DIFF
--- a/src/verifai/features/__init__.py
+++ b/src/verifai/features/__init__.py
@@ -1,1 +1,2 @@
 from .features import *
+from .features import _PrecomputedSample
--- a/src/verifai/features/__init__.py
+++ b/src/verifai/features/__init__.py
@@ -1,2 +1,1 @@
 from .features import *
-from .features import _PrecomputedSample
--- a/src/verifai/features/features.py
+++ b/src/verifai/features/features.py
@@ -1013,7 +1013,7 @@ class FeatureSpace:
         })
     """
 
-    def __init__(self, features, distanceMetric=None, timeBound=None):
+    def __init__(self, features, distanceMetric=None, timeBound=0):
         self.namedFeatures = tuple(sorted(features.items(), key=lambda i: i[0]))
         self.featureNamed = OrderedDict(self.namedFeatures)
         self.features = tuple(self.featureNamed.values())
@@ -1023,12 +1023,15 @@ class FeatureSpace:
         self.dynamicFeatureNamed = OrderedDict({name: feat for name, feat in self.featureNamed.items()
                                                if isinstance(feat, TimeSeriesFeature)})
 
-        self.makePoint = namedtuple('SpacePoint', self.featureNamed)
-        self.makeStaticPoint = namedtuple('SpacePoint', self.staticFeatureNamed)
-        self.makeDynamicPoint = namedtuple('SpacePoint', self.dynamicFeatureNamed)
+        # self.makePoint = namedtuple('SpacePoint', self.featureNamed)
+        self.makeStaticPoint = namedtuple('StaticSpacePoint', self.staticFeatureNamed)
+        self.makeDynamicPoint = namedtuple('DynamicSpacePoint', self.dynamicFeatureNamed)
 
         self.distanceMetric = distanceMetric
         self.timeBound = timeBound
+
+        if len(self.dynamicFeatureNamed) > 0 and self.timeBound == 0:
+            raise RuntimeError("FeatureSpace which includes TimeSeriesFeature has no timeBound.")
 
     @cached_property
     def domains(self):
@@ -1041,8 +1044,6 @@ class FeatureSpace:
         of all features. If any Features are TimeSeriesFeatures then they are
         expanded to the a max of timeBound.
         """
-        assert len(self.dynamicFeatureNamed) == 0 or self.timeBound is not None
-
         fixedDomains = {}
         lengthDomains = {}
         variableDomains = {}
@@ -1081,11 +1082,14 @@ class FeatureSpace:
         """Flatten a point in this space. See Domain.flatten.
 
         If fixedDimension is True, the point is flattened out as if all feature
-        lists had their maximum lengths, with None as a placeholder. This means
-        that all points in the space will flatten to the same length.
+        lists had their maximum lengths and time steps, with None as a placeholder.
+        This means that all points in the space will flatten to the same length.
         """
+        from verifai.samplers.feature_sampler import Sample
+        assert isinstance(point, Sample)
+
         flattened = []
-        for feature, value in zip(self.features, point):
+        for feature, value in zip(self.staticFeatureNamed.values(), point.staticSample):
             domain = feature.domain
             if feature.lengthDomain:
                 length = len(value)
@@ -1099,6 +1103,39 @@ class FeatureSpace:
                         flattened.append(None)
             else:
                 domain.flattenOnto(value, flattened)
+
+        flattened.append(len(point.dynamicSampleHistory))
+
+        for feature_i, feature in enumerate(self.dynamicFeatureNamed.values()):
+            if feature.lengthDomain:
+                length = len(value)
+            else:
+                length = None
+
+            flattened.append(length)
+            
+            for dynamic_point in point.dynamicSampleHistory:
+                value = dynamic_point[feature_i]
+                domain = feature.domain
+
+                if length is None:
+                    domain.flattenOnto(value, flattened)
+                else:
+                    fixedDomain = feature.fixedDomains(self.timeBound)[length]
+                    fixedDomain.flattenOnto(value, flattened)
+                    if fixedDimension:
+                        sizePerElt = domain.flattenedDimension
+                        needed = (feature.maxLength - length) * sizePerElt
+                        for i in range(needed):
+                            flattened.append(None)
+
+            if fixedDimension:
+                needed = (self.space.timeBound - len(point.dynamicSampleHistory)) * length * sizePerElt
+                flattened += [None for _ in range(needed)]
+
+        flattened_point = tuple(flattened)
+        if fixedDimension:
+            assert len(flattened_point) == self.fixedFlattenedDimension
         return tuple(flattened)
 
     @cached_property
@@ -1108,13 +1145,15 @@ class FeatureSpace:
         Also an upper bound on the length of the vector returned by flatten
         by default, when fixedDimension=False."""
         dim = 0
+        dim += 1 # Timesteps
         for feature in self.features:
             domain = feature.domain
+            timeMult = self.timeBound if isinstance(feature, TimeSeriesFeature) else 1
             if feature.lengthDomain:
                 dim += 1    # dimension storing length of the feature list
-                dim += feature.maxLength * domain.flattenedDimension
+                dim += timeMult * feature.maxLength * domain.flattenedDimension
             else:
-                dim += domain.flattenedDimension
+                dim += timeMult * domain.flattenedDimension
         return dim
 
     def meaningOfFlatCoordinate(self, index, pointName='point'):
@@ -1125,26 +1164,53 @@ class FeatureSpace:
         have different meaning depending on the lengths of feature lists.
         """
         assert 0 <= index < self.fixedFlattenedDimension
-        for name, feature in self.namedFeatures:
+        for name, feature in self.staticFeatureNamed.items():
             domain = feature.domain
             if feature.lengthDomain:
                 if index == 0:
-                    return f'len({pointName}.{name})'
+                    return f'len({pointName}.staticSample.{name})'
                 else:
                     index -= 1
                     elem = index // domain.flattenedDimension
                     if elem < feature.maxLength:
                         subIndex = index % domain.flattenedDimension
-                        subPoint = f'{pointName}.{name}[{elem}]'
+                        subPoint = f'{pointName}.staticSample.{name}[{elem}]'
                         return domain.meaningOfFlatCoordinate(subIndex,
                             pointName=subPoint)
                     index -= feature.maxLength * domain.flattenedDimension
             else:
                 if index < domain.flattenedDimension:
-                    subPoint = f'{pointName}.{name}'
+                    subPoint = f'{pointName}.staticSample.{name}'
                     return domain.meaningOfFlatCoordinate(index,
                                                           pointName=subPoint)
                 index -= domain.flattenedDimension
+
+        if index == 0:
+            return f'len({pointName}.dynamicSampleHistory)'
+        index -= 1
+
+        for name, feature in self.dynamicFeatureNamed.items():
+            domain = feature.domain
+            for time_i in range(self.timeBound):
+                if feature.lengthDomain:
+                    if index == 0:
+                        return f'len({pointName}.dynamicSampleHistory[{time_i}].{name})'
+                    else:
+                        index -= 1
+                        elem = index // domain.flattenedDimension
+                        if elem < feature.maxLength:
+                            subIndex = index % domain.flattenedDimension
+                            subPoint = f'{pointName}.dynamicSampleHistory[{time_i}].{name}[{elem}]'
+                            return domain.meaningOfFlatCoordinate(subIndex,
+                                pointName=subPoint)
+                        index -= feature.maxLength * domain.flattenedDimension
+                else:
+                    if index < domain.flattenedDimension:
+                        subPoint = f'{pointName}.dynamicSampleHistory[{time_i}].{name}'
+                        return domain.meaningOfFlatCoordinate(index,
+                                                            pointName=subPoint)
+                    index -= domain.flattenedDimension
+
         raise RuntimeError('impossible index arithmetic')
 
     def pandasIndexForFlatCoordinate(self, index):
@@ -1152,6 +1218,7 @@ class FeatureSpace:
 
         See meaningOfFlatCoordinate, and Domain.pandasIndexForFlatCoordinate.
         """
+        #TODO Update
         assert 0 <= index < self.fixedFlattenedDimension
         for name, feature in self.namedFeatures:
             domain = feature.domain
@@ -1179,7 +1246,7 @@ class FeatureSpace:
         See meaningOfFlatCoordinate, and Domain.coordinateIsNumerical.
         """
         assert 0 <= index < self.fixedFlattenedDimension
-        for name, feature in self.namedFeatures:
+        for name, feature in self.staticFeatureNamed.items():
             domain = feature.domain
             if feature.lengthDomain:
                 if index == 0:
@@ -1195,26 +1262,86 @@ class FeatureSpace:
                 if index < domain.flattenedDimension:
                     return domain.coordinateIsNumerical(index)
                 index -= domain.flattenedDimension
+
+        if index == 0:
+            return True
+        index -= 1
+
+        for name, feature in self.dynamicFeatureNamed.items():
+            domain = feature.domain
+            for time_i in range(self.timeBound):
+                if feature.lengthDomain:
+                    if index == 0:
+                        return True
+                    else:
+                        index -= 1
+                        elem = index // domain.flattenedDimension
+                        if elem < feature.maxLength:
+                            subIndex = index % domain.flattenedDimension
+                            return domain.coordinateIsNumerical(subIndex)
+                        index -= feature.maxLength * domain.flattenedDimension
+                else:
+                    if index < domain.flattenedDimension:
+                        return domain.coordinateIsNumerical(index)
+                    index -= domain.flattenedDimension
+
         raise RuntimeError('impossible index arithmetic')
 
     def unflatten(self, coords, fixedDimension=False):
         """Unflatten a tuple of coordinates to a point in this space."""
-        values = []
+        from verifai.samplers.feature_sampler import LateFeatureSample
+
+        staticValues = []
         iterator = iter(coords)
-        for feature in self.features:
+
+        for feature in self.staticFeatureNamed.values():
             domain = feature.domain
             if feature.lengthDomain:
                 length = next(iterator)
                 fixedDomain = feature.fixedDomains(self.timeBound)[length]
-                values.append(fixedDomain.unflattenIterator(iterator))
+                staticValues.append(fixedDomain.unflattenIterator(iterator))
                 if fixedDimension:      # consume padding
                     sizePerElt = domain.flattenedDimension
                     needed = (feature.maxLength - length) * sizePerElt
                     for i in range(needed):
                         next(iterator)
             else:
-                values.append(domain.unflattenIterator(iterator))
-        return self.makePoint(*values)
+                staticValues.append(domain.unflattenIterator(iterator))
+
+        staticSample = self.makeStaticPoint(*staticValues)
+
+        timeSteps = next(iterator)
+
+        dynamicValuesList = [[] for _ in range(timeSteps)]
+
+        for feature in self.dynamicFeatureNamed.values():
+            domain = feature.domain
+            length = next(iterator)
+            
+            for time_i in range(timeSteps):
+
+
+                if length is None:
+                    dynamicValuesList[time_i].append(domain.unflattenIterator(iterator))
+                else:
+                    fixedDomain = feature.fixedDomains(self.timeBound)[length]
+                    dynamicValuesList[time_i].append(fixedDomain.unflattenIterator(iterator))
+                    if fixedDimension:      # consume padding
+                        sizePerElt = domain.flattenedDimension
+                        needed = (feature.maxLength - length) * sizePerElt
+                        for i in range(needed):
+                            next(iterator)
+
+            if fixedDimension:
+                needed = (self.space.timeBound - timeSteps) * length * sizePerElt
+                for i in range(needed):
+                    next(iterator)
+
+        dynamicSampleList = [self.makeDynamicPoint(*dynamicValues) for dynamicValues in dynamicValuesList]
+
+        updateCallback = lambda rho: None
+
+        return LateFeatureSample(space=None, staticSample=staticSample, dynamicSampleList=dynamicSampleList, updateCallback=updateCallback)
 
     def __repr__(self):
         rep = f'FeatureSpace({self.featureNamed}'
--- a/src/verifai/features/features.py
+++ b/src/verifai/features/features.py
@@ -964,19 +964,26 @@ class Feature:
         else:
             return self.distanceMetric(valueA, valueB)
 
-    @cached_property
-    def fixedDomains(self):
+    @staticmethod
+    def _timeExpandDomain(domain, timeBound):
+        return domain
+
+    def fixedDomains(self, timeBound):
         """Return the fixed-length Domains associated with this feature."""
+        timeExpandedDomain = self._timeExpandDomain(self.domain, timeBound)
+
         if not self.lengthDomain:
-            return self.domain
-        domains = {}
-        for length in self.lengthDomain:
-            length = length[0]
-            domains[length] = Array(self.domain, (length,))
+            domains = timeExpandedDomain
+        else:
+            domains = {}
+            for length in self.lengthDomain:
+                length = length[0]
+                domains[length] = Array(timeExpandedDomain, (length,))
+        
         return domains
 
     def __repr__(self):
-        rep = f'Feature({self.domain}'
+        rep = f'{self.__class__.__name__}({self.domain}'
         if self.distribution is not None:
             rep += f', distribution={self.distribution}'
         if self.lengthDomain is not None:
@@ -986,6 +993,11 @@ class Feature:
         if self.distanceMetric is not None:
             rep += f', distanceMetric={self.distanceMetric}'
         return rep + ')'
+
+class TimeSeriesFeature(Feature):
+    @staticmethod
+    def _timeExpandDomain(domain, timeBound):
+        return Array(domain, (timeBound,))
 
 ### Feature spaces
 
@@ -1001,30 +1013,43 @@ class FeatureSpace:
         })
     """
 
-    def __init__(self, features, distanceMetric=None):
+    def __init__(self, features, distanceMetric=None, timeBound=None):
         self.namedFeatures = tuple(sorted(features.items(), key=lambda i: i[0]))
         self.featureNamed = OrderedDict(self.namedFeatures)
         self.features = tuple(self.featureNamed.values())
-        self.makePoint = namedtuple('SpacePoint', self.featureNamed.keys())
+
+        self.staticFeatureNames = OrderedDict({name: feat for name, feat in self.featureNamed.items()
+                                               if not isinstance(feat, TimeSeriesFeature)})
+        self.dynamicFeatureNames = OrderedDict({name: feat for name, feat in self.featureNamed.items()
+                                               if isinstance(feat, TimeSeriesFeature)})
+
+        self.makePoint = namedtuple('SpacePoint', self.featureNamed)
+        self.makeStaticPoint = namedtuple('SpacePoint', self.staticFeatureNames)
+        self.makeDynamicPoint = namedtuple('SpacePoint', self.dynamicFeatureNames)
+
         self.distanceMetric = distanceMetric
+        self.timeBound = timeBound
 
     @cached_property
     def domains(self):
-        """Return the domain or domains associated with this space.
+        """Return the expanded domain or domains associated with this space.
 
         Returns a pair consisting of the Domain of all lengths of feature
         lists, plus a dict mapping each (flattened) point in that Domain to the
         corresponding Domain of other features. If the FeatureSpace has no
         feature lists, then returns (None, dom) where dom is the fixed Domain
-        of all features.
+        of all features. If any Features are TimeSeriesFeatures then they are
+        expanded to the a max of timeBound.
         """
+        assert self.timeBound is not None
+
         fixedDomains = {}
         lengthDomains = {}
         variableDomains = {}
         for name, feature in self.namedFeatures:
             if feature.lengthDomain:
                 lengthDomains[name] = feature.lengthDomain
-                variableDomains[name] = feature.fixedDomains
+                variableDomains[name] = feature.fixedDomains(self.timeBound)
             else:
                 fixedDomains[name] = feature.domain
         if len(lengthDomains) == 0:
@@ -1065,7 +1090,7 @@ class FeatureSpace:
             if feature.lengthDomain:
                 length = len(value)
                 flattened.append(length)
-                fixedDomain = feature.fixedDomains[length]
+                fixedDomain = feature.fixedDomains(self.timeBound)[length]
                 fixedDomain.flattenOnto(value, flattened)
                 if fixedDimension:      # add padding to maximum length
                     sizePerElt = domain.flattenedDimension
@@ -1180,7 +1205,7 @@ class FeatureSpace:
             domain = feature.domain
             if feature.lengthDomain:
                 length = next(iterator)
-                fixedDomain = feature.fixedDomains[length]
+                fixedDomain = feature.fixedDomains(self.timeBound)[length]
                 values.append(fixedDomain.unflattenIterator(iterator))
                 if fixedDimension:      # consume padding
                     sizePerElt = domain.flattenedDimension
--- a/src/verifai/features/features.py
+++ b/src/verifai/features/features.py
@@ -1018,14 +1018,14 @@ class FeatureSpace:
         self.featureNamed = OrderedDict(self.namedFeatures)
         self.features = tuple(self.featureNamed.values())
 
-        self.staticFeatureNames = OrderedDict({name: feat for name, feat in self.featureNamed.items()
+        self.staticFeatureNamed = OrderedDict({name: feat for name, feat in self.featureNamed.items()
                                                if not isinstance(feat, TimeSeriesFeature)})
-        self.dynamicFeatureNames = OrderedDict({name: feat for name, feat in self.featureNamed.items()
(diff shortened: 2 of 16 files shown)


src/verifai/features/features.py
```diff
@@ -1001,32 +1013,46 @@ class FeatureSpace:
...
         self.featureNamed = OrderedDict(self.namedFeatures)
         self.features = tuple(self.featureNamed.values())
-        self.makePoint = namedtuple('SpacePoint', self.featureNamed.keys())
+
+        self.staticFeatureNamed = OrderedDict({name: feat for name, feat in self.featureNamed.items()
+                                               if not isinstance(feat, TimeSeriesFeature)})
+        self.dynamicFeatureNamed = OrderedDict({name: feat for name, feat in self.featureNamed.items()
+                                               if isinstance(feat, TimeSeriesFeature)})
+
+        # self.makePoint = namedtuple('SpacePoint', self.featureNamed)
```
dfremont: Presumably left by accident.

src/verifai/features/features.py
```diff
@@ -1001,32 +1013,46 @@ class FeatureSpace:
...
-        """Return the domain or domains associated with this space.
+        """Return the expanded domain or domains associated with this space.
 
         Returns a pair consisting of the Domain of all lengths of feature
         lists, plus a dict mapping each (flattened) point in that Domain to the
         corresponding Domain of other features. If the FeatureSpace has no
         feature lists, then returns (None, dom) where dom is the fixed Domain
-        of all features.
+        of all features. If any Features are TimeSeriesFeatures then they are
+        expanded to the a max of timeBound.
```
dfremont:
```suggestion
        expanded to a max of timeBound.
```

src/verifai/features/features.py
```diff
@@ -1001,32 +1013,46 @@ class FeatureSpace:
         })
     """
 
-    def __init__(self, features, distanceMetric=None):
+    def __init__(self, features, distanceMetric=None, timeBound=0):
```
dfremont, line 1114: Should describe `timeBound` in the docstring. (What it does, and the type -- presumably needs to be an integer number of time steps.)

src/verifai/features/features.py
```diff
@@ -987,6 +994,11 @@ def __repr__(self):
             rep += f', distanceMetric={self.distanceMetric}'
         return rep + ')'
 
+class TimeSeriesFeature(Feature):
```
dfremont, line 998: Need docstring. (This class is mentioned in other places.)

src/verifai/features/features.py
```diff
@@ -1056,16 +1082,19 @@ def flatten(self, point, fixedDimension=False):
         """Flatten a point in this space. See Domain.flatten.
 
         If fixedDimension is True, the point is flattened out as if all feature
-        lists had their maximum lengths, with None as a placeholder. This means
-        that all points in the space will flatten to the same length.
+        lists had their maximum lengths and time steps, with None as a placeholder.
+        This means that all points in the space will flatten to the same length.
         """
+        from verifai.samplers.feature_sampler import Sample
```
dfremont: I think it would make more sense for `Sample` to be defined here, alongside `FeatureSpace` (it's a point in the feature space after all). This would also get rid of the cyclic dependency.

src/verifai/features/features.py
```diff
@@ -1074,6 +1103,39 @@ def flatten(self, point, fixedDimension=False):
                         flattened.append(None)
             else:
                 domain.flattenOnto(value, flattened)
+
+        flattened.append(len(point.dynamicSampleHistory))
+
+        for feature_i, feature in enumerate(self.dynamicFeatureNamed.values()):
+            if feature.lengthDomain:
+                length = len(value)
```
dfremont: Surely using `value` from the previous loop was not intended. Is there a test exercising this code with features of differing lengths?

src/verifai/features/features.py
```diff
@@ -1074,6 +1103,39 @@ def flatten(self, point, fixedDimension=False):
...
+                if length is None:
+                    domain.flattenOnto(value, flattened)
+                else:
+                    fixedDomain = feature.fixedDomains(self.timeBound)[length]
+                    fixedDomain.flattenOnto(value, flattened)
+                    if fixedDimension:
+                        sizePerElt = domain.flattenedDimension
+                        needed = (feature.maxLength - length) * sizePerElt
+                        for i in range(needed):
+                            flattened.append(None)
```
dfremont: Can we pull this code out into a helper function (can be local, doesn't need to be a method on the class) and use it in the loop for static features too? Most of the code is identical.

dfremont: (We can leave this to later, it's not that important)

src/verifai/features/features.py
```diff
@@ -1074,6 +1103,39 @@ def flatten(self, point, fixedDimension=False):
                         flattened.append(None)
             else:
                 domain.flattenOnto(value, flattened)
+
+        flattened.append(len(point.dynamicSampleHistory))
```
dfremont: If the feature space does not use time series at all, can we omit this? Then the flattened vectors won't change for existing code, which is nice for backwards compatibility as well as eliminating a redundant column in the error table. (You'd have to remove the +1 in `fixedFlattenedDimension` too in this case.)

src/verifai/features/features.py
```diff
@@ -1083,13 +1145,15 @@ def fixedFlattenedDimension(self):
         Also an upper bound on the length of the vector returned by flatten
         by default, when fixedDimension=False."""
         dim = 0
+        dim += 1 # Timesteps
         for feature in self.features:
             domain = feature.domain
+            timeMult = self.timeBound if isinstance(feature, TimeSeriesFeature) else 1
```
dfremont, line 1254: A thought probably best left for later: maybe we should move all the flattening logic into new `Feature._flatten` and `Feature._flattenedDimension` methods. Then this kind of case split could be handled by overrides in `TimeSeriesFeature` instead of needing to have special cases here.

src/verifai/features/features.py
```diff
@@ -1100,33 +1164,61 @@ def meaningOfFlatCoordinate(self, index, pointName='point'):
         have different meaning depending on the lengths of feature lists.
         """
         assert 0 <= index < self.fixedFlattenedDimension
-        for name, feature in self.namedFeatures:
+        for name, feature in self.staticFeatureNamed.items():
             domain = feature.domain
             if feature.lengthDomain:
                 if index == 0:
-                    return f'len({pointName}.{name})'
+                    return f'len({pointName}.staticSample.{name})'
```
dfremont: This cluttering of the column names is somewhat unfortunate. Features need to have distinct names anyway, so can we avoid this somehow? It would be nice if we could just have something like the following, which would preserve the names of the old columns:

- `point.blah` for plain old features (as before);
- `point.blah[index]` for feature lists (as before);
- `point.blah[timestep]` for time series features;
- `point.blah[index][timestep]` for time series feature lists.

If the problem is avoiding features with the same name as methods on `Sample`, I think banning a few names like `update` as names of features (maybe only when time series are actually being used) would not be so bad.

src/verifai/features/features.py
```diff
@@ -1100,33 +1164,61 @@ def meaningOfFlatCoordinate(self, index, pointName='point'):
...
+                    index -= domain.flattenedDimension
+
         raise RuntimeError('impossible index arithmetic')
 
     def pandasIndexForFlatCoordinate(self, index):
         """Pandas index of a coordinate of a flattened point in this space.
 
         See meaningOfFlatCoordinate, and Domain.pandasIndexForFlatCoordinate.
         """
+        #TODO Update
```
dfremont: Honestly I can't remember what this method is supposed to be for: I can't find any usages of it in my code at all. If updating it to support time series is complex, we could remove it. But since it is documented, we should not leave it in a broken state.

src/verifai/samplers/domain_sampler.py
```diff
@@ -47,22 +47,24 @@ def update(self, sample, info, rho):
...
-        """Generate the next sample, given feedback from the last sample.
-
-        This exists only for backwards-compatibility. It has been replaced by
-        the getSample and update APIs.
-        """
-        if self.last_sample is not None:
-            self.update(self.last_sample, self.last_info, feedback)
-        self.last_sample, self.last_info = self.getSample()
-        return self.last_sample
+    # TODO: Deprecate
```
dfremont: If it's hard to make this still work, let's just remove it.

src/verifai/samplers/feature_sampler.py
```diff
@@ -21,9 +23,32 @@
 from verifai.samplers.simulated_annealing import SimulatedAnnealingSampler
 from verifai.samplers.grid_sampler import GridSampler
 
+class Sample(ABC):
```
dfremont: Need docstrings for class and public methods.

src/verifai/samplers/feature_sampler.py
```diff
@@ -211,41 +251,73 @@ class LateFeatureSampler(FeatureSampler):
...
             self.domainSamplers = {
                 point: makeDomainSampler(domain)
                 for point, domain in fixedDomains.items()
             }
-            self.lastLength = None
+
+        self.id_metadata_dict = {}
+        self._last_id = 0
+
+    def get_info_id(self, info, length, sample):
```
dfremont: This is a private, internal method, right?

src/verifai/samplers/feature_sampler.py
```diff
@@ -211,41 +251,73 @@ class LateFeatureSampler(FeatureSampler):
...
+                point_dict = {}
+
+                for f, val in dynamic_features:
+                    if not self.space.featureNamed[f].lengthDomain:
+                        point_dict[f] = val[t]
+                    else:
+                        feat_list = []
+                        for l in range(len(val)):
+                            feat_list.append(val[l][t])
+                        point_dict[f] = tuple(feat_list)
```
dfremont:
```suggestion
                        point_dict[f] = tuple(v[t] for v in val)
```

src/verifai/samplers/feature_sampler.py
```diff
@@ -194,12 +204,42 @@ def restoreFromFile(path):
...
+        self._i += 1
+
+        return dynamic_sample
+
+    def update(self, rho):
+        return self._updateCallback(rho)
+
 class LateFeatureSampler(FeatureSampler):
     """FeatureSampler that works by first sampling only lengths of feature
     lists, then sampling from the resulting fixed-dimensional Domain.
```
dfremont: Need to update docstring to explain handling of time series. (Also please put it in the correct format with a 1-line summary, then a blank line, then the detailed description -- looks like I messed that up originally.)

src/verifai/samplers/feature_sampler.py
```diff
@@ -211,41 +251,73 @@ class LateFeatureSampler(FeatureSampler):
...
         else:
             self.lengthDomain = lengthDomain
             self.lengthSampler = makeLengthSampler(lengthDomain)
             self.domainSamplers = {
                 point: makeDomainSampler(domain)
                 for point, domain in fixedDomains.items()
             }
-            self.lastLength = None
+
+        self.id_metadata_dict = {}
```
dfremont: Should be private.

src/verifai/features/features.py
```diff
@@ -1170,26 +1262,86 @@ def coordinateIsNumerical(self, index):
...
+                    if index < domain.flattenedDimension:
+                        return domain.coordinateIsNumerical(index)
+                    index -= domain.flattenedDimension
+
         raise RuntimeError('impossible index arithmetic')
 
     def unflatten(self, coords, fixedDimension=False):
         """Unflatten a tuple of coordinates to a point in this space."""
-        values = []
+        from verifai.samplers.feature_sampler import LateFeatureSample
```
dfremont: This suggests some connection to `LateFeatureSampler`, when there isn't one. If the essential thing is that a `LateFeatureSample` is completely determined at construction time and doesn't actually call out to a sampler for the dynamic part, you could call it `CompleteSample` or `ReplaySample` or `LookupSample` or something and put it next to `Sample` in `features.py` (or a new `samples.py` if that would be helpful).

src/verifai/samplers/scenic_sampler.py
```diff
@@ -223,6 +223,22 @@ def spaceForScenario(scenario, ignoredProperties):
...
+        super().__init__(space)
+        self._staticSample = staticSample
+        self._updateCallback = updateCallback
+
+    @property
+    def staticSample(self):
+        return self._staticSample
+
+    def _getDynamicSample(self, info):
+        raise RuntimeError("ScenicSample does not support dynamic sampling.")
```
dfremont: `ScenicSample` is a private class, right? In which case people will be confused to see it referenced here if they were just misusing `ScenicSampler`.

src/verifai/samplers/scenic_sampler.py
```diff
@@ -314,7 +362,7 @@ def pointForScene(self, scene):
             params[param] = pointForValue(subdom, scene.params[originalName])
         paramPoint = paramDomain.makePoint(**params)
 
-        return self.space.makePoint(objects=objPoint, params=paramPoint)
+        return self.space.makeStaticPoint(objects=objPoint, params=paramPoint)
```
dfremont: For consistency I suppose this should return a `ScenicSample`, although that will break existing code. If we leave the current behavior, then I'm not sure how to make the docstring accurate, since I don't think we describe the concept of "the static part of the feature space" anywhere.

src/verifai/server.py
```diff
@@ -144,6 +144,8 @@ def __init__(self, sampling_data, monitor, options={}):
                 sampler_params=params
             )
 
+        if (self.sample_space.dynamicFeatureNamed) != 0:
```
dfremont: Rather than peeking inside like this, how about defining a flag like `FeatureSpace.hasTimeSeries` and using it here and in a couple other places where you do a check like this?

tests/scenic/test_scenic.py
```diff
@@ -179,3 +179,56 @@ def test_driving_dynamic(pathToLocalFile):
...
+        n_iters=3,
+        save_error_table=False,
+        save_safe_table=False,
+    )
+    server_options = DotMap(maxSteps=2, verbosity=3)
+    falsifier = generic_falsifier(sampler=sampler,
+                                  falsifier_params=falsifier_params,
+                                  server_class=ScenicServer,
+                                  server_options=server_options)
+    falsifier.run_falsifier()
```
dfremont, line 202: Let's either extend this test or add a new one where you actually check that the time series is being sampled separately at each time step.

tests/test_features.py
```diff
@@ -10,39 +10,39 @@ def test_fs_flatten():
...
-        point = sampler.nextSample()
+        point = sampler.getSample()
         flat = space.flatten(point, fixedDimension=True)
         assert type(flat) is tuple
-        assert len(flat) == 4
-        bLen = len(point.b)
-        assert eval(space.meaningOfFlatCoordinate(0)) == point.a[0]
+        assert len(flat) == 5
+        bLen = len(point.staticSample.b)
+        assert eval(space.meaningOfFlatCoordinate(0, pointName='point')) == point.staticSample.a[0]
```
dfremont: Shouldn't need to set `pointName` here and below: the default is `point` already.

tests/test_samplers.py
```diff
@@ -3,9 +3,46 @@
...
+        'c': TimeSeriesFeature(Box((2,5))),
+        'd': TimeSeriesFeature(Box((5,6)), lengthDomain=DiscreteBox((0,2)))
+    },
+    timeBound=10)
+    sampler = FeatureSampler.randomSamplerFor(space)
+
+    sample = sampler.getSample()
+    static_point = sample.staticSample
+
+    static_dict = static_point._asdict()
```
dfremont: I think the `_asdict()` doesn't clarify anything. We intentionally use named tuples to simplify the notation, so you can just write `static_point.a` for example.

tests/utils.py
```diff
@@ -19,18 +19,27 @@ def checkSaveRestore(sampler, tmpdir, iterations=1):
     feedback = None
     for i in range(iterations):
         sampler.saveToFile(path)
-        sample1 = sampler.nextSample(feedback)
-        sample2 = sampler.nextSample(-1)
+        # sample1 = sampler.nextSample(feedback)
+        # sample2 = sampler.nextSample(-1)
```
dfremont: Left accidentally (some more below too).

dfremont (changes requested):
Looks good overall. Details are in comments above, but my main suggestion is to try to preserve the simple way to access features from the sampled point (`point.myFeature` instead of `point.staticSample.myFeature`). If we can do that, then most of the changes you had to make to the existing tests shouldn't be necessary.

src/verifai/features/features.py
Eric-Vin: Good catch, and added a test as well!

src/verifai/features/features.py
Eric-Vin: I implemented this, but with `point.blah[timestep][index]` instead of `point.blah[index][timestep]` as that matches the current implementation better. Let me know if you want me to make changes!

src/verifai/features/features.py
```diff
@@ -995,15 +996,112 @@ def __repr__(self):
...
+    @abstractmethod
+    def update(self, rho):
+        pass
+
+    def __getattr__(self, attr):
+        space = super().__getattribute__("space")
+        if attr in space.staticFeatureNamed:
+            return getattr(self.staticSample, attr)
+        elif attr in space.dynamicFeatureNamed:
+            class DynamicFeatureHelper:
```
dfremont, line 1042: Any reason not to also implement `__len__` and inherit from `collections.abc.Sequence`?

src/verifai/features/features.py
```diff
@@ -987,11 +995,113 @@ def __repr__(self):
             rep += f', distanceMetric={self.distanceMetric}'
         return rep + ')'
 
-### Feature spaces
+class TimeSeriesFeature(Feature):
+    """A feature with a value at each timesetep of a simulation."""
```
dfremont, line 999:
```suggestion
    """A feature with a value at each timestep of a simulation."""
```

src/verifai/samplers/feature_sampler.py
```diff
@@ -149,29 +151,14 @@ def makeDomainSampler(domain):
...
-        This function exists only for backwards compatibility. It has been
-        superceded by the `getSample` and `update` APIs.
-        """
-        if self.last_sample is not None:
-            self.update(self.last_sample, self.last_info, feedback)
-        self.last_sample, self.last_info = self.getSample()
-        return self.last_sample
+    @abstractmethod
+    def getSample(self):
+        """Returns a `Sample` object"""
```
dfremont, line 160:
```suggestion
        """Generate a `Sample`."""
```

src/verifai/samplers/feature_sampler.py
```diff
@@ -194,15 +181,22 @@ def restoreFromFile(path):
...
 
+
 class LateFeatureSampler(FeatureSampler):
-    """FeatureSampler that works by first sampling only lengths of feature
-    lists, then sampling from the resulting fixed-dimensional Domain.
+    """ FeatureSampler that greedily samples a CompleteSample.
+    
+    FeatureSampler works as follows:
+        1. Sample lengths of feature lists.
+        2. Expand TimeSeriesFeatures into flattened features with of length
```
dfremont, line 197:
```suggestion
        2. Expand TimeSeriesFeatures into flattened features of length
```

src/verifai/samplers/feature_sampler.py
```diff
@@ -194,15 +181,22 @@ def restoreFromFile(path):
...
+           space.timeBound.
+        2. Sample from the resulting fixed-dimensional Domains.
```
dfremont, line 199:
```suggestion
        3. Sample from the resulting fixed-dimensional Domains.
```

src/verifai/features/features.py
```diff
@@ -1170,29 +1393,99 @@ def coordinateIsNumerical(self, index):
...
+        else:
+            duration = 0
+            dynamicSampleList = []
+            dynamicSampleLengths = {}
+
+        updateCallback = lambda rho: None
+
+        sample = CompleteSample(space=self, staticSample=staticSample, dynamicSampleList=dynamicSampleList, updateCallback=updateCallback,dynamicSampleLengths=dynamicSampleLengths)
+
+        for _ in range(duration):
```
Eric-Vin, line 1481: Replace with `unroll` method? @dfremont feel free to suggest a different name.

src/verifai/features/features.py
```diff
@@ -987,11 +995,113 @@ def __repr__(self):
...
+                    if i > len(self.dynamicSampleHistory):
+                        raise ValueError("Attempting to access dynamic sample value that has not been sampled.")
+                    return getattr(self.dynamicSampleHistory[i], self.attr)
+
+            return DynamicFeatureHelper(self.dynamicSampleHistory, attr)
+        else:
+            return super().__getattr__(attr)
+
+
+class CompleteSample(Sample):
```
dfremont, line 1057: This is an implementation detail, so let's make this class private. Users should only need to know about the `Sample` API.

src/verifai/features/features.py
```diff
@@ -1170,29 +1393,99 @@ def coordinateIsNumerical(self, index):
...
+            duration = 0
+            dynamicSampleList = []
+            dynamicSampleLengths = {}
+
+        updateCallback = lambda rho: None
+
+        sample = CompleteSample(space=self, staticSample=staticSample, dynamicSampleList=dynamicSampleList, updateCallback=updateCallback,dynamicSampleLengths=dynamicSampleLengths)
+
+        for _ in range(duration):
+            sample.getDynamicSample()
```
dfremont, line 1482: As we discussed, this is probably better handled by an option to `CompleteSample`.

src/verifai/samplers/feature_sampler.py
```diff
@@ -194,15 +181,22 @@ def restoreFromFile(path):
...
         except TerminationException:
             return
 
+
 class LateFeatureSampler(FeatureSampler):
-    """FeatureSampler that works by first sampling only lengths of feature
-    lists, then sampling from the resulting fixed-dimensional Domain.
+    """ FeatureSampler that greedily samples a CompleteSample.
+    
+    FeatureSampler works as follows:
```
dfremont, line 195:
```suggestion
    LateFeatureSampler works as follows:
```

src/verifai/samplers/feature_sampler.py
```diff
@@ -194,15 +181,22 @@ def restoreFromFile(path):
...
+                rho = yield sample
+                sample.update(rho)
         except TerminationException:
             return
 
+
 class LateFeatureSampler(FeatureSampler):
-    """FeatureSampler that works by first sampling only lengths of feature
-    lists, then sampling from the resulting fixed-dimensional Domain.
+    """ FeatureSampler that greedily samples a CompleteSample.
```
dfremont, line 193: Need to reword this in a clearer way that doesn't refer to the private `CompleteSample`.

src/verifai/samplers/feature_sampler.py
```diff
@@ -211,41 +205,76 @@ class LateFeatureSampler(FeatureSampler):
...
+        info = (info1, info2)
+        
+        sample_id = self._get_info_id(info, length, domainPoint)
+        update_callback = lambda rho: self.update(sample_id, rho)
+
+        # Make static points and iterable over dynamic points
+        static_features = [v for v in domainPoint._asdict().items()
+                            if v[0] in self.space.staticFeatureNamed]
+        dynamic_features = [v for v in domainPoint._asdict().items()
+                            if v[0] not in self.space.staticFeatureNamed]
```
dfremont, line 245: This seems like a lot of work to be doing every sample. Can't we precompute the static and dynamic feature names? Alternatively, maybe it would simplify the code to split the domain into two domains, one for the static features and one for the dynamic features (i.e. `FeatureSpace.staticDomains` and `FeatureSpace.dynamicDomains`).

src/verifai/samplers/feature_sampler.py
```diff
@@ -211,41 +205,76 @@ class LateFeatureSampler(FeatureSampler):
...
+        static_point = self.space.makeStaticPoint(*[v[1] for v in static_features])
+
+        dynamic_points = []
+        if any(isinstance(f, TimeSeriesFeature) for f in self.space.features):
```
dfremont, line 249:
```suggestion
        if self.space.hasTimeSeries:
```

src/verifai/samplers/feature_sampler.py
```diff
@@ -211,41 +205,76 @@ class LateFeatureSampler(FeatureSampler):
...
+            for t in range(self.space.timeBound):
+                point_dict = {}
```
dfremont, line 251: Can't you just use a list rather than a dict?

src/verifai/samplers/scenic_sampler.py
```diff
@@ -262,25 +280,44 @@ def fromScenario(cls, path, maxIterations=None,
               e.g. ``params`` to override global parameters or ``model`` to set the
               :term:`world model`.
         """
+        if "params" not in kwargs:
+            kwargs["params"] = {}
+
+        kwargs["params"]["timeBound"] = maxSteps if maxSteps else 0
```
dfremont, line 286:
```suggestion
        params = kwargs.setdefault("params", {})
        params["timeBound"] = maxSteps if maxSteps else 0
```
Also, arguably if `maxSteps` is not specified here but the Scenic program defines `timeBound` (which will need to be documented in the Scenic docs, in the External Parameters section), we should use the value in the program. For that you could use `maxSteps or params.get("timeBound", 0)`.

src/verifai/samplers/scenic_sampler.py
```diff
@@ -262,25 +280,44 @@ def fromScenario(cls, path, maxIterations=None,
...
         scenario = scenic.scenarioFromFile(path, **kwargs)
         return cls(scenario, maxIterations=maxIterations,
                    ignoredProperties=ignoredProperties)
 
     @classmethod
-    def fromScenicCode(cls, code, maxIterations=None,
+    def fromScenicCode(cls, code, maxIterations=None, maxSteps=None,
                        ignoredProperties=None, **kwargs):
         """As above, but given a Scenic program as a string."""
+        if "params" not in kwargs:
```
dfremont, line 296: As above.

tests/scenic/test_scenic.py
```diff
@@ -179,3 +179,56 @@ def test_driving_dynamic(pathToLocalFile):
...
+behavior TestBehavior():
+    while True:
+        foo.getSample()
+        foo.getSample()
+        wait
+ego = new Object with behavior TestBehavior()
+"""
+
+def test_double_time_series_access():
+    with pytest.raises(RuntimeError):
```
dfremont, line 216: This is rather broad. Can you check for a specific exception message?

tests/scenic/test_scenic.py
dfremont: You still aren't looking at the actual sampled values, right? My point was to check that the value at time 0 is not the same as the value at time 1.

tests/scenic/scenic_driving_behavior.scenic
```diff
@@ -0,0 +1,19 @@
+param map = localPath('Town01.xodr')
+param carla_map = 'Town01'
```
dfremont, line 2: Not needed, since this is running in Newtonian.

tests/scenic/scenic_driving_behavior.scenic
```diff
@@ -0,0 +1,19 @@
...
+
+model scenic.domains.driving.model
```
dfremont, line 4: Why not just specify Newtonian here and then you don't need to when compiling the scenario?

tests/scenic/test_scenic.py
dfremont: Oops, I see now that you're doing it in the scenario itself. Sorry!

tests/test_grid.py
```diff
@@ -14,7 +14,9 @@ def test_grid():
     dict_samples = defaultdict(int)
     while True:
         try:
-            sample = sampler.nextSample()
+            sample = sampler.getSample()
+            sample.update(None)
+            sample = sample
```
dfremont, line 19: ?

tests/test_grid.py
```diff
@@ -46,7 +48,9 @@ def f(sample):
     y_samples = []
 
     for i in range(21):
-        sample = sampler.nextSample()
+        sample = sampler.getSample()
+        sample.update(None)
+        sample = sample
```
dfremont, line 53: ?

tests/test_grid.py
```diff
@@ -59,6 +63,6 @@ def test_grid_non_standardizable():
         'b': Feature(FilteredDomain(Box([0,1]), lambda x: x[0] > 0.5))
     })
     sampler = FeatureSampler.gridSamplerFor(space)
-    samples = list(sampler)
+    samples = [s for s in sampler]
```
dfremont, line 66: Why this change? It's equivalent, right?

tests/test_samplers.py
```diff
@@ -3,9 +3,41 @@
...
 from verifai.samplers import RandomSampler, FeatureSampler
 
+def test_feature_sampling():
+    space = FeatureSpace({
+        'a': Feature(DiscreteBox([0, 12])),
+        'b': Feature(Box((0, 1)), lengthDomain=DiscreteBox((0, 2))),
+        'c': TimeSeriesFeature(Box((2,5))),
+        'd': TimeSeriesFeature(Box((5,6)), lengthDomain=DiscreteBox((0,2)))
+    },
+    timeBound=10)
```
dfremont, line 16: Indentation here is confusing.

tests/test_samplers.py
```diff
@@ -72,8 +103,8 @@ def check(samples):
         assert any(len(sample.b) == 1 for sample in samples)
         assert any(len(sample.b) == 2 for sample in samples)
 
-    check([sampler.nextSample() for i in range(100)])
-    check(list(itertools.islice(sampler, 100)))
+    check([sampler.getSample() for i in range(100)])
+    check(list(s for s in itertools.islice(sampler, 100)))
```
dfremont, line 107: Redundant comprehension.

tests/test_features.py
Eric-Vin, line 1: Update tests to not reference `dynamicSampleHistory`

src/verifai/features/features.py
```diff
@@ -987,11 +995,113 @@ def __repr__(self):
...
+    """A sample from a feature space, containing static points and able to generate dynamic points.
+    
+    Args:
+        space (FeatureSpace): The feature space this Sample was sampled from.
+        dynamicSampleLengths (dict): A dictionary containing the lengths of each dynamic feature
+            with a length domain.
+    """
+    def __init__(self, space, dynamicSampleLengths):
+        self.space = space
+        self.dynamicSampleHistory = []
```
Eric-Vin: Make private

src/verifai/features/features.py
Eric-Vin: Rename `PrecomputedSample` and add a new `CompletedSample` representing a finalized sample that is to be used for flattening/unflattening. Also make sure we support hashing, equality, etc...

src/verifai/features/features.py
```diff
@@ -987,11 +995,113 @@ def __repr__(self):
...
+    def _getDynamicSample(self, info):
+        pass
+
+    def getDynamicSample(self, info=None):
+        sample = self._getDynamicSample(info)
+        self.dynamicSampleHistory.append(sample)
+        return sample
+
+    @abstractmethod
+    def update(self, rho):
```
Eric-Vin: Rename to `complete` to match `CompletedSample`.

src/verifai/samplers/feature_sampler.py
Eric-Vin: I don't think I fully follow here, what's the operation you're referring to here?

src/verifai/samplers/scenic_sampler.py
Eric-Vin: Good suggestion. I added your fix and I'll leave the note open as a reminder.

src/verifai/features/features.py
```diff
@@ -987,11 +996,158 @@ def __repr__(self):
...
+            return getattr(self.staticSample, attr)
+        elif attr in space.dynamicFeatureNamed:
+            class DynamicFeatureHelper(Sequence):
+                def __init__(self, dynamicSampleHistory, attr):
+                    self._dynamicSamples = dynamicSampleHistory
+                    self.attr = attr
+
+                def __getitem__(self, i):
+                    if i > len(self._dynamicSamples):
+                        raise ValueError("Attempting to access dynamic sample value that has not been sampled.")
```
dfremont:
```suggestion
                        raise IndexError("Attempting to access dynamic sample value that has not been sampled.")
```

src/verifai/features/features.py
```diff
@@ -987,11 +996,158 @@ def __repr__(self):
...
+                def __len__(self):
+                    return len(self._dynamicSamples)
+
+            return DynamicFeatureHelper(self._dynamicSamples, attr)
+        else:
+            return super().__getattribute__(attr)
+
+
+class Sample(_SampleBase):
+    """A sample from a feature space, containing static points and able to generate dynamic points.
```
dfremont:
```suggestion
    """A sample from a feature space, containing a static point and able to generate dynamic points.
```

src/verifai/features/features.py
```diff
@@ -987,11 +996,158 @@ def __repr__(self):
...
+        return self._dynamicSamples
+
+    def __eq__(self, other):
+        if not isinstance(other, CompletedSample):
+            return False
+        
+        return self.staticSample == other.staticSample and tuple(self.dynamicSamples) == tuple(other.dynamicSamples)
+
+    def __hash__(self):
+        return hash(self.staticSample, tuple(self.dynamicSamples))
```
dfremont: I guess there's no test for this, since `hash` only takes 1 argument.

src/verifai/features/features.py
```diff
@@ -987,11 +996,158 @@ def __repr__(self):
...
+class CompletedSample(_SampleBase):
+    """A completed sample from a feature space, containing the static point the history of sampled dynamic points.
+    
+    This class is returned when calling `complete` on a `Sample`, and is useful for storage, comparisons, hashing, etc...
+    """
+
+    def __init__(self, staticSample, dynamicSamples, space, dynamicSampleLengths):
+        super().__init__(space, dynamicSampleLengths)
+        self._staticSample = staticSample
+        self._dynamicSamples = dynamicSamples
```
dfremont:
```suggestion
        self._dynamicSamples = tuple(dynamicSamples)
```
Then you can remove the calls to `tuple` in `__eq__` and `__hash__` below.

src/verifai/features/features.py
```diff
@@ -1001,32 +1157,47 @@ class FeatureSpace:
...
+        self.hasTimeSeries = len(self.dynamicFeatureNamed) > 0
+
+        self.makeStaticPoint = namedtuple('StaticSpacePoint', self.staticFeatureNamed)
+        self.makeDynamicPoint = namedtuple('DynamicSpacePoint', self.dynamicFeatureNamed)
+
         self.distanceMetric = distanceMetric
+        self.timeBound = timeBound
+
+        if len(self.dynamicFeatureNamed) > 0 and self.timeBound == 0:
+            raise RuntimeError("FeatureSpace which includes TimeSeriesFeature has no timeBound.")
```
dfremont:
```suggestion
            raise ValueError("must specify timeBound when creating a FeatureSpace with a TimeSeriesFeature")
```

src/verifai/features/__init__.py
```diff
@@ -1 +1,2 @@
 from .features import *
+from .features import _PrecomputedSample
```
dfremont: I don't think we want people to get this if they do `from verifai.features import *`, since it's an internal class. I would instead change our own imports to `from verifai.features.features import _PrecomputedSample`.

src/verifai/samplers/feature_sampler.py
```diff
@@ -211,41 +205,74 @@ class LateFeatureSampler(FeatureSampler):
...
+
+        domainPoint, info2 = self.domainSamplers[length].getSample()
+        info = (info1, info2)
+        
+        sample_id = self._get_info_id(info, length, domainPoint)
+        complete_callback = lambda rho: self.update(sample_id, rho)
+
+        # Make static points and iterable over dynamic points
+        static_features = [(k, domainPoint._asdict()[k]) for k in self.space.staticFeatureNamed.keys()]
+        dynamic_features = [(k, domainPoint._asdict()[k]) for k in self.space.dynamicFeatureNamed.keys()]
```
dfremont:
```suggestion
        static_features = [(k, domainPoint._asdict()[k]) for k in self.space.staticFeatureNamed]
        dynamic_features = [(k, domainPoint._asdict()[k]) for k in self.space.dynamicFeatureNamed]
```

tests/test_features.py
```diff
@@ -87,12 +87,53 @@ def test_fs_flatten_fixed_dimension2():
...
+        assert flat[1] == duration
+        for t in range(duration):
+            offset = 2*t
+            bLen = len(point.dynamicSamples[t].b)
+            assert flat[2] == bLen
+            assert eval(space.meaningOfFlatCoordinate(2)) == bLen
+            if bLen < 1:
+                assert flat[offset+3] is None
+            else:
+                assert eval(space.meaningOfFlatCoordinate(offset+3)) == point.dynamicSamples[t].b[0][0]
```
dfremont, line 119: I think we don't have any tests for the alternate way of spelling this, namely `point.b[t][0][0]`. I'd suggest adding a separate test, since it isn't specific to flattening. We should test it on both complete and incomplete samples.

dfremont (changes requested):
Looks good overall, just a few remaining minor comments.

dfremont (approved):
Thanks a lot Eric!

PR #59, opened 2025-12-02
Add dynamic rulebook code and example

kevinchang73:


beyazit-y (approved):

examples/dynamic_rulebook/maps/Town05.xodr
dfremont, line 1: Is the example really specific to Town05? If not, can we change it to use Town01 so that it's not necessary to include another map in the repo? (Town01 is already under the `tests` folder)

examples/dynamic_rulebook/multi_01/util/multi_01_analyze_diversity.py
```diff
@@ -0,0 +1,61 @@
...
+blocking_car_dist = []
+bypass_dist = []
+
+ego_speed_max = []
+dist_threshold_max = []
+blocking_car_dist_max = []
+bypass_dist_max = []
+
+for file in all_files:
+    infile = open(directory+'/'+file, 'r')
```
dfremont, line 25: Use a `with` statement so that the file is properly closed.

Also, hard-coding `/` will not work on Windows: use `pathlib` or `os.path` to manipulate paths. (Actually in this particular case `glob.glob` would probably be best, since then you don't need to manually use `endswith` and `startswith` above.)

examples/dynamic_rulebook/multi_01/util/multi_01_analyze_diversity.py
```diff
@@ -0,0 +1,61 @@
+import sys
+import matplotlib.pyplot as plt
+import numpy as np
+import os
```
dfremont, line 4: In all files, please group imports into 3 blocks separated by 1 blank line: Python stdlib, external packages, local imports from VerifAI itself.

examples/dynamic_rulebook/multi_01/util/multi_01_collect_result.py
```diff
@@ -0,0 +1,144 @@
...
+                counterexample_type_2[curr_source][tuple(1*np.array([val3[1]]))] += 1
+            else:
+                counterexample_type_2[curr_source][tuple(1*np.array([val3[1]]))] = 1
+
+            if order == '-1':
+                curr_source = curr_source + 1 if curr_source < 2 else 0
+
+print('Error weights')
+print('segment 0:')
+for i in range(1):
```
dfremont, line 119: Why these unnecessary loops? If this code is copied from a template, why not expose the general template as a helper function which you can import here?

examples/dynamic_rulebook/multi_01/multi_01.scenic
```diff
@@ -0,0 +1,92 @@
...
+
+DIST_TO_INTERSECTION = 15
+TERM_DIST = 40
+
+#################################
+# AGENT BEHAVIORS               #
+#################################
+
+behavior EgoBehavior(path):
+    current_lane = network.laneAt(self)
```
dfremont, line 34:
```suggestion
    current_lane = self.lane
```
It doesn't look like you're handling the case when the car is not in a lane, so I've used the `lane` property here, which will reject the simulation if the car isn't in a lane. If you actually want `current_lane` to be `None` in that case, use the `_lane` property instead (documentation [here](https://docs.scenic-lang.org/en/latest/modules/scenic.domains.driving.model.html#scenic.domains.driving.model.DrivingObject.lane)).

examples/dynamic_rulebook/multi_01/multi_01.scenic
```diff
@@ -0,0 +1,92 @@
...
+    laneChangeCompleted = False
+    bypassed = False
+    try:
+        do FollowLaneBehavior(globalParameters.EGO_SPEED, laneToFollow=current_lane)
+    interrupt when (distance to blockingCar) < globalParameters.DIST_THRESHOLD and not laneChangeCompleted:
+        do LaneChangeBehavior(path, is_oppositeTraffic=True, target_speed=globalParameters.EGO_SPEED)
+        do FollowLaneBehavior(globalParameters.EGO_SPEED, is_oppositeTraffic=True) until (distance to blockingCar) > globalParameters.BYPASS_DIST
+        laneChangeCompleted = True
+    interrupt when (blockingCar can see ego) and (distance to blockingCar) > globalParameters.BYPASS_DIST and not bypassed:
+        current_laneSection = network.laneSectionAt(self)
```
dfremont, line 44:
```suggestion
        current_laneSection = self.laneSection
```

examples/dynamic_rulebook/multi_01/multi_01.scenic
```diff
@@ -0,0 +1,92 @@
...
+ego = new Car at spawnPt,
+    with blueprint MODEL,
+    with behavior EgoBehavior(leftLaneSec)
+    
+blockingCar = new Car following roadDirection from ego for globalParameters.BLOCKING_CAR_DIST,
+            with blueprint MODEL,
+            with viewAngle 90 deg
+
+require (distance from blockingCar to intersection) > DIST_TO_INTERSECTION
+terminate when (distance to spawnPt) > TERM_DIST
```
dfremont, line 83:
```suggestion
require distance from blockingCar to intersection > DIST_TO_INTERSECTION
terminate when distance to spawnPt > TERM_DIST
```

examples/dynamic_rulebook/multi_01/multi_01.scenic
```diff
@@ -0,0 +1,92 @@
...
+
+#################################
+# RECORDING                     #
+#################################
+
+record initial (initLaneSec.polygon.exterior.coords) as initLaneCoords
```
dfremont, line 89: Likewise, the parser should be able to handle having no parentheses here.

examples/dynamic_rulebook/multi_01/multi_01_spec.py
```diff
@@ -0,0 +1,19 @@
+import numpy as np
+
+def rule0(simulation, indices): # safe distance to obstacle
```
dfremont, line 3: Did you decide not to implement my suggestion for a decorator to reduce the boilerplate here?

examples/dynamic_rulebook/multi_02/multi_02.scenic
```diff
@@ -0,0 +1,125 @@
...
+BYPASS_DIST = 10
+SAFE_DIST = 10
+INIT_DIST = 40
+TERM_DIST = 80
+
+#################################
+# AGENT BEHAVIORS               #
+#################################
+
+behavior DecelerateBehavior(brake):
```
dfremont, line 38: What's the purpose of having a behavior that only takes 1 action? Why not just inline the body?

examples/dynamic_rulebook/multi_03/multi_03.scenic
```diff
@@ -0,0 +1,177 @@
...
+#################################
+# AGENT BEHAVIORS               #
+#################################
+
+behavior EgoBehavior(trajectory):
+    flag = True
+    try:
+        do FollowTrajectoryBehavior(target_speed=globalParameters.EGO_SPEED, trajectory=trajectory)
+        do FollowLaneBehavior(target_speed=globalParameters.ADV_SPEED)
+    interrupt when withinDistanceToAnyObjs(self, SAFETY_DIST) and (ped in network.drivableRegion) and flag:
```
dfremont, line 50: Is the intended behavior here that once this interrupt fires the first time, it can never fire again because the flag is set to false? That's a little surprising, so it might be good to explain what the ego is trying to do either here or in the docstring for the whole scenario (which is just a stub).

examples/dynamic_rulebook/multi_03/multi_03.scenic
```diff
@@ -0,0 +1,177 @@
...
+    with blueprint MODEL_ADV,
+    with behavior Adv3Behavior(adv3Trajectory)
+
+adv4 = new Car at ego offset by -10 @ 85,
+    with blueprint MODEL_ADV,
+    with behavior Adv3Behavior(adv4Trajectory)
+
+ped = new Pedestrian at pedSpawnPt,
+    facing toward pedEndPt,
+    with regionContainedIn None,
```
dfremont, line 156: This is the default (meaning the pedestrian can be anywhere in the workspace). Did you mean `everywhere`?

examples/dynamic_rulebook/multi_inter_left/multi_inter_left.scenic
```diff
@@ -0,0 +1,137 @@
+"""
+TITLE: Verifai 2.0 Left Turn
```
dfremont, line 2:
```suggestion
TITLE: VerifAI 3.0 Left Turn
```

examples/dynamic_rulebook/multi_inter_left/multi_inter_left.scenic
```diff
@@ -0,0 +1,137 @@
...
+record (distance from ego to network.drivableRegion) as egoDistToDrivableRegion
+record (distance from ego to egoInitLane.group) as egoDistToEgoInitLane
+record (distance from ego to egoManeuver.endLane.group) as egoDistToEgoEndLane
+record (distance from ego to ego.lane.centerline) as egoDistToEgoLaneCenterline
+record (distance from ego to intersection) as egoDistToIntersection
+
+record (distance from ego to adv1) as egoDistToAdv1
+record (distance to egoSpawnPt) as egoDistToEgoSpawnPt
+
+record ego._boundingPolygon as egoPoly
```
dfremont, line 130: I don't think we should be using an internal, undocumented property in our examples. You can use `ego.boundingPolygon` to get a `PolygonalRegion`, which should be fine for whatever you need to do with it.

Actually having said that, it doesn't seem that these polygons are used anywhere. So I would remove them.

examples/dynamic_rulebook/multi.py
```diff
@@ -0,0 +1,179 @@
...
+                   experiment_name=None, max_time=None, n_iters=None, max_steps=300):
+    if not os.path.exists(output_dir):
+        os.mkdir(output_dir)
+    paths = []
+    if os.path.isdir(path):
+        for root, _, files in os.walk(path):
+            for name in files:
+                fname = os.path.join(root, name)
+                if os.path.splitext(fname)[1] == '.scenic':
+                    paths.append(fname)
```
dfremont, line 53: I think something like `glob.glob(f"{path}/**/*.scenic")` would work just as well here and be a lot easier to read.

examples/dynamic_rulebook/multi.py
```diff
@@ -0,0 +1,179 @@
...
+    else:
+        paths = [path]
+    for p in paths:
+        falsifier = run_experiment(p, rulebook=rulebook, 
+        parallel=parallel, model=model, sampler_type=sampler_type, headless=headless,
+        num_workers=num_workers, max_time=max_time, n_iters=n_iters, max_steps=max_steps)
```
dfremont, line 59: Confusing indentation

examples/dynamic_rulebook/multi.py
```diff
@@ -0,0 +1,179 @@
...
+"""
+Runs a single falsification experiment.
+
+Arguments:
+    path: Path to Scenic script to be run.
+    parallel: Whether or not to enable parallelism.
+    model: Which simulator model to use (e.g. scenic.simulators.newtonian.driving_model)
+    sampler_type: Which VerifAI sampelr to use (e.g. halton, scenic, ce, mab, etc.)
+    headless: Whether or not to display each simulation.
+    num_workers: Number of parallel workers. Only used if parallel is true.
```
dfremont, line 86: Missing some arguments here

examples/dynamic_rulebook/multi.py
```diff
@@ -0,0 +1,179 @@
...
+    falsifier = falsifier_class(monitor=rb, ## modified
+                                sampler_type=s_type, 
+                                sampler=sampler, 
+                                falsifier_params=falsifier_params,
+                                server_options=server_options,
+                                server_class=ScenicServer)
+    print(f'(multi.py) Sampler type: {falsifier.sampler_type}')
+    
+    # Run falsification
+    t0 = time.time()
```
dfremont, line 124: All timings should use `time.monotonic()` so as not to be affected by changes to the system clock.

examples/dynamic_rulebook/multi.py
```diff
@@ -0,0 +1,179 @@
...
+    if not parallel:
+        print(f'(multi.py) Sampling time: {falsifier.total_sample_time}')
+        print(f'(multi.py) Simulation time: {falsifier.total_simulate_time}')
+    print(f'(multi.py) Confidence interval: {falsifier.get_confidence_interval()}')
+    return falsifier
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--scenic-path', '-sp', type=str, default='uberCrashNewton.scenic',
+    help='Path to Scenic script')
```
dfremont, line 140: Confusing indentation (please go through all files for this kind of thing, or run a tool like Black)

examples/dynamic_rulebook/README.md
```diff
@@ -0,0 +1,248 @@
...
+First, follow the instructions in the [VerifAI documentation](https://verifai.readthedocs.io/en/latest/installation.html) to create a virtual environment and install VerifAI:
+
+```bash
+python3 -m venv venv_verifai
+source venv_verifai/bin/activate
+
+git clone https://github.com/BerkeleyLearnVerify/VerifAI
+cd VerifAI
+python -m pip install --upgrade pip
+python -m pip install -e .
```
dfremont, line 16:
```suggestion
python -m pip install -e ".[examples]"
```

examples/dynamic_rulebook/README.md
```diff
@@ -0,0 +1,248 @@
...
+```
+
+Then, for this example, we adopt the [Metadrive simulator](https://metadriverse.github.io/metadrive/) as the backend simulator. To install Metadrive, run the following command:
```
dfremont, line 19: You should just add MetaDrive and sumolib to the `examples` extra instead of requiring a separate `pip` invocation (which can break dependency constraints).

examples/dynamic_rulebook/README.md
```diff
@@ -0,0 +1,248 @@
...
+```bash
+python -m pip install "metadrive-simulator @ git+https://github.com/metadriverse/metadrive.git@main"
+python -m pip install "sumolib >= 1.21.0"
+```
+
+As there exists a dependency conflict on the `progressbar` package between VerifAI and Metadrive, we need to uninstall the `progressbar` package and reinstall the `progressbar2` package:
+
+```bash
+python -m pip uninstall progressbar
+python -m pip install --force-reinstall progressbar2==3.55.0
```
dfremont, line 30: This is not a good idea. Is it hard to fix our code to use `progressbar2`? I would hope it would be a simple change.

examples/dynamic_rulebook/README.md
```diff
@@ -0,0 +1,248 @@
...
+```
+
+## Running the Examples
+
+We provide six different scenarios in the `examples/dynamic_rulebook` folder, and you can run any of them by modifying and executing the `run_multi_dynamic.sh` script.
+
+```bash linenums="1"
```
dfremont, line 37: I think quoting the entire script here is confusing. I'm not sure you need to quote any of it, but if you think that would be helpful, just quote the specific part people should modify.

.gitignore
```diff
@@ -120,3 +120,5 @@ venv.bak/
 dmypy.json
 
 *.cproject
+
+examples/dynamic_rulebook/*/outputs/
```
dfremont, line 124: Let's move this to a new `.gitignore` file in the `examples/dynamic_rulebook` directory, so that it will keep working even if we move that folder later.

examples/dynamic_rulebook/run_multi_dynamic.sh
```diff
@@ -0,0 +1,37 @@
...
+result_file="result_${scenario}_${sampler_type}_${sampler_idx}_${use_dynamic_rulebook}.txt"
+csv_file="result_${scenario}_${sampler_type}_${sampler_idx}_${use_dynamic_rulebook}"
+
+rm $scenario/outputs/$log_file
+rm $scenario/outputs/$result_file
+rm $scenario/outputs/$csv_file.*csv
+rm $scenario/outputs/$csv_file\_scatter.png
+if [ "$use_dynamic_rulebook" = true ]; then
+
+    for seed in $(seq 0 1);
```
dfremont, line 20: Why are we running this twice with different fixed seeds?

src/verifai/rulebook.py
```diff
@@ -0,0 +1,185 @@
...
+
+    def visit_FunctionDef(self, node):
+        self.functions.append(node)
+
+class rulebook(ABC):
+    priority_graphs = {}
+    using_sampler = -1
+    verbosity = 1
+    exploration_ratio = 2.0
+    using_continuous = False
```
dfremont, line 22: Surely all of these should be attributes of the rulebook object, not global state attached to the class? What if somebody wants to instantiate multiple rulebooks with different parameters?

src/verifai/rulebook.py
```diff
@@ -0,0 +1,185 @@
...
+from verifai.monitor import specification_monitor
+
+class FunctionVisitor(ast.NodeVisitor):
+    def __init__(self):
+        self.functions = []
+
+    def visit_FunctionDef(self, node):
+        self.functions.append(node)
+
+class rulebook(ABC):
```
dfremont, line 17:
```suggestion
class Rulebook(ABC):
```

src/verifai/samplers/dynamic_rulebook_ce.py
```diff
@@ -0,0 +1,138 @@
...
+                                                                RandomSampler)
+            for subsampler in self.split_samplers[id].samplers:
+                if isinstance(subsampler, ContinuousDynamicCESampler):
+                    subsampler.set_graph(priority_graph)
+                elif isinstance(subsampler, DiscreteDynamicCESampler):
+                    assert True
+                else:
+                    assert isinstance(subsampler, RandomSampler)
+        if not sorted(list(self.split_samplers.keys())) == list(range(len(rulebook.priority_graphs))):
+            raise ValueError('Priority graph IDs should be in order and start from 0')
```
dfremont, line 47: Shouldn't this be an assertion? We're the ones deciding the structure of the rulebook object and how we're putting together the samplers here, not the user.

src/verifai/samplers/dynamic_rulebook_ce.py
```diff
@@ -0,0 +1,138 @@
...
+            (lambda d: d.standardizedIntervals, self.disc_ce)
+        )
+        self.split_samplers = {}
+        for id, priority_graph in rulebook.priority_graphs.items():
+            self.split_samplers[id] = SplitSampler.fromPartition(domain,
+                                                                partition,
+                                                                RandomSampler)
+            for subsampler in self.split_samplers[id].samplers:
+                if isinstance(subsampler, ContinuousDynamicCESampler):
+                    subsampler.set_graph(priority_graph)
```
dfremont, line 41: Setting the graph in a second step seems unnecessary. Can't you just pass it in when creating the sampler originally? (in the lambdas above, which you'd have to move inside the loop)

src/verifai/samplers/dynamic_rulebook_ce.py
```diff
@@ -0,0 +1,138 @@
...
+            idx = self.sampler_idx % self.num_segs
+        else:
+            idx = self.using_sampler
+        return self.split_samplers[idx].getSample()
+
+    def update(self, sample, info, rhos):
+        # Update each sampler based on the corresponding segment
+        try:
+            iter(rhos)
+        except:
```
dfremont, line 65: Don't use bare `except:`, which will catch `KeyboardInterrupt`. If you really want to catch any ordinary exception you can use `except Exception:`, but in this case you seem to just be checking whether `rho` is iterable or not. Doesn't it have to be a tuple since we're doing multi-objective falsification?

src/verifai/falsifier.py
```diff
@@ -159,6 +163,12 @@ def run_falsifier(self):
                     break
                 if self.verbosity >= 2:
                     print("Sample no: ", i, "\nSample: ", sample, "\nRho: ", rho)
+                if self.dynamic and self.verbosity >= 1:
```
dfremont, line 166: This should only happen at verbosity 2, in which case we just printed `rho` above, so why do we need this?

src/verifai/falsifier.py
```diff
@@ -176,15 +186,23 @@ def run_falsifier(self):
...
-                    self.populate_error_table(sample, rho)
-                ce_num = ce_num + 1
-                if ce_num >= self.ce_num_max:
-                    break
-            elif self.save_safe_table:
-                self.populate_error_table(sample, rho, error=False)
+            ce = False
+            if self.dynamic:
+                for r in rho:
+                    self.populate_error_table(sample, r)
```
dfremont, line 192: Why are we calling this multiple times? There should only be one row in the error table for each sample/simulation. The different elements of `rho` should become different columns, as in the old multi-objective falsification code, right?

src/verifai/scenic_server.py
```diff
@@ -51,11 +51,18 @@ def __init__(self, sampling_data, monitor, options={}):
...
     def evaluate_sample(self, sample):
         scene = self.sampler.lastScene
         assert scene
         result = self._simulate(scene)
+        if self.dynamic:
+            while result is None:
+                sample = self.get_sample(1)
+                scene = self.sampler.lastScene
+                assert scene
+                result = self._simulate(scene)
```
dfremont, line 65: What is the purpose of this code? We should not be running more than one simulation per sample. If the simulation is rejected, the sampler should receive the rejection feedback as in the following lines.

src/verifai/rulebook.py
```diff
@@ -0,0 +1,185 @@
...
+
+        function_visitor = FunctionVisitor()
+        function_visitor.visit(tree)
+
+        self.functions = {}
+        for function_node in function_visitor.functions:
+            function_name = function_node.name
+            function_code = compile(ast.Module(body=[function_node], type_ignores=[]), '<string>', 'exec')
+            exec(function_code)
+            self.functions[function_name] = locals()[function_name]
```
dfremont, line 54: Rather than all this rigamarole of compiling each function separately (which incidentally would break if the function refers to a constant defined at the top level of the module, for example), why not just `exec` the entire file into a new namespace and pull out all the functions? `runpy.run_path` might work for this too.

src/verifai/falsifier.py
dfremont: OK, if I understand correctly, in the dynamic case you've changed `rho` to be a list of lists, one per segment. But it still doesn't make sense to have multiple rows for a single simulation, and the `rho_0` column for example will mean different things in different rows, since the first rule varies depending on which segment was active. We'll need to have a discussion on Tuesday about what the error table should look like.

src/verifai/samplers/dynamic_rulebook_ce.py
```diff
@@ -0,0 +1,138 @@
...
+        self.buckets = buckets # 1*d, each element specifies the number of buckets in that dimension
+        self.dist = dist # N*d
+        self.alpha = alpha
+        self.thres = thres
+        self.current_sample = None
+
+    def getVector(self):
+        return self.generateSample()
+    
+    def generateSample(self):
```
dfremont, line 105: What's the point of separating this method from `getVector`, which is the actual API required for a `BoxSampler`?

src/verifai/samplers/dynamic_rulebook_ce.py
```diff
@@ -0,0 +1,138 @@
...
+        assert rho is not None
+        self.update_dist_from_multi(vector, info, rho)
+    
+    def update_dist_from_multi(self, sample, info, rho):
+        try:
+            iter(rho)
+        except:
+            return
+        if len(rho) != self.num_properties:
+            return
```
dfremont, line 123: Returning silently doesn't seem like the right thing to do. Shouldn't this be an assertion?

src/verifai/samplers/dynamic_rulebook_ce.py
```diff
@@ -0,0 +1,138 @@
...
+        
+        is_ce = True
+        for node in self.priority_graph.nodes:
+            if rho[node] >= self.thres[node]:
+                is_ce = False
+                break
```
dfremont, line 129: Is this the intended logic? It seems to say you only consider the sample to be a counterexample if it violates *all* of the specifications in the graph.

src/verifai/samplers/dynamic_rulebook_ce.py
```diff
@@ -0,0 +1,138 @@
...
+        
+        if not is_ce:
+            return
+        for row, b in zip(self.dist, info):
+            row *= self.alpha
+            row[b] += 1 - self.alpha
+
+class DiscreteDynamicCESampler(DiscreteCrossEntropySampler):
```
dfremont, line 137: Did you test this? I don't think it will work, since `DiscreteCrossEntropySampler` does not expect `rho` to be a list.

src/verifai/error_table.py
```diff
@@ -38,15 +38,15 @@ def update_column_names(self, column_names):
...
+    def update_error_table(self, sample, rho, is_multi=False):
         sample = self.space.flatten(sample, fixedDimension=True)
         sample_dict = {}
         for k, v in zip(self.table.columns, list(sample)):
             if np.any(np.array(sample) == None):
                 locs = np.where(np.array(sample) == None)
                 self.ignore_locs = self.ignore_locs + list(locs[0])
             sample_dict[k] = float(v) if self.column_type[k] and v is not None else v
-        if isinstance(rho, (list, tuple)):
+        if is_multi or isinstance(rho, (list, tuple)):
```
dfremont, line 49: Why is this change needed? If `rho` is not a list/tuple, won't iterating over it fail? Or are you passing some new kind of iterable in for `rho`?

dfremont (changes requested):
Hi Kai-Chun, thanks for making the PR. I haven't finished reviewing the new samplers and the rulebook code itself, but won't get to it till later so I wanted to send what I have now. Besides my detailed comments above, some big-picture points:
 
* There are important changes to the server and error table that don't seem necessary and which don't make sense to me. Please see my comments, and we can discuss in detail on Tuesday.
* We need test cases for the new functionality. I'm pretty sure some of the code won't work (e.g. discrete domains with the new DCE sampler), which would come out if there were tests.
* There is a lot of repeated code across the examples which I think could be substantially reduced, e.g. using decorators for rules that operate uniformly on trajectories (see our previous discussion), a helper function to segment trajectories based on when specified records become true, and a common base for the `collect_result` scripts. This isn't just to save lines: the code will be easier to understand if it has a simpler structure and the common parts are broken out into functions with informative names.
* The description of the system in the README is great and very comprehensive, but I think it's long enough to be hard to navigate (it combines basic instructions on running the example with technical details about the APIs). Since we need to add a new page to the online documentation about this feature anyway, I would suggest moving the current README into the documentation (either as one big page with different sections [probably replacing the current Multi-Objective Falsification page], or possibly multiple pages) and having only the essentials in the README itself. You could include installation instructions, how to run the example, and a brief summary of what the different files are for, pointing to the documentation for details.

examples/dynamic_rulebook/maps/Town05.xodr
kevinchang73: It seems that Town01 doesn't work for some scenarios. I'll keep using Town05.

examples/dynamic_rulebook/run_multi_dynamic.sh
kevinchang73: This aligns with the experiments in the original RV '24 paper, where we ran the same scenario using different random seeds.

src/verifai/samplers/dynamic_rulebook_ce.py
kevinchang73: Since the graph IDs come from the input `*.graph` files, they are considered user inputs.

kevinchang73:
I’ve addressed all the feedback by updating the code and replying to the specific threads. The only remaining item is migrating the README to the documentation. I believe a few other sections of the docs also require updates. Should we hold off and discuss the overall documentation structure first, or would you prefer I migrate the README immediately?

Please let me know if you have any other suggestions. Thanks!

dfremont:
Thanks, Kai-Chun! We can skip making the new documentation page for now, but please extract the instructions for running your part of the case study and send them to Kai so he can include them in the artifact README.

examples/dynamic_rulebook/multi_02/multi_02.scenic
```diff
@@ -0,0 +1,122 @@
...
+#################################
+
+record ego.lane is initLane or ego.lane is not adv2.lane as egoIsInInitLane
+record adv2.lane is initLane as adv2IsInInitLane # start evaluation only when adv2 reaches another lane
+record adv3.lane is initLane as adv3IsInInitLane # start evaluation only when adv3 reaches another lane
+
+record ego._boundingPolygon as egoPoly
+record adv1._boundingPolygon as adv1Poly
+record adv2._boundingPolygon as adv2Poly
+record adv3._boundingPolygon as adv3Poly
```
dfremont: Per my comment on one of the other scenarios, let's get rid of these if they aren't being used. (The polygons below are better since they don't use undocumented properties, but you could also get rid of them if they aren't used.)

examples/dynamic_rulebook/multi_03/util/multi_03_collect_result.py
```diff
@@ -0,0 +1,252 @@
...
+            line = lines[i + 1].strip().split(" ")
+            val1 = []
+            val_print = []
+            for s in line:
+                if s != "":
+                    val1.append(float(s) < 0)
+                    val_print.append(float(s))
+            assert len(val1) == 4, "Invalid length of rho"
+            result_count_0[curr_source].append(
+                val1[0] * 8 + val1[1] * 4 + val1[3] * 2 + val1[2] * 1
```
dfremont, line 37: Is the index sequence really supposed to be 0, 1, 3, 2 here? (and in a few places below)

examples/dynamic_rulebook/multi_03/util/multi_03_collect_result.py
```diff
@@ -0,0 +1,252 @@
...
+            )
+            if (
+                tuple(1 * np.array([val1[0], val1[1], val1[3], val1[2]]))
```
dfremont: Once you fix the indices, isn't this just equivalent to `tuple(val1)`?

examples/dynamic_rulebook/multi_03/util/multi_03_collect_result.py
```diff
@@ -0,0 +1,252 @@
...
+                + val2[5] * 2
+                + val2[6] * 1
+            )
+            if (
+                tuple(
+                    1
+                    * np.array(
+                        [val2[0], val2[1], val2[2], val2[3], val2[4], val2[5], val2[6]]
+                    )
+                )
```
dfremont: This is just `tuple(val2)`. Is this code being generated by some other script?

examples/dynamic_rulebook/multi.py
```diff
@@ -0,0 +1,286 @@
...
+    return falsifier
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "--scenic-path",
+        "-sp",
+        type=str,
+        default="uberCrashNewton.scenic",
```
dfremont: This default doesn't make sense since there is no such file in this directory.

src/verifai/samplers/dynamic_rulebook_ce.py
```diff
@@ -0,0 +1,141 @@
...
+            thres=self.thres,
+            priority_graph=priority_graph,
+        )
+        self.split_samplers = {}
+        for id, priority_graph in rb.priority_graphs.items():
+            self.split_samplers[id] = self.cont_ce(domain, priority_graph)
+        if not sorted(list(self.split_samplers.keys())) == list(
+            range(len(rb.priority_graphs))
+        ):
+            raise ValueError("Priority graph IDs should be in order and start from 0")
```
dfremont: I don't understand the condition here: since you sort the keys, you're not actually checking whether they're in order prior to sorting. I think you're just checking whether the keys are consecutive integers starting at 0.

Anyway, is it actually possible for the condition to be violated? If so, it seems like a bug in the Rulebook code (which should set up whatever numbering you need), so this should be an assertion rather than an exception. If it is possible for the user to construct an invalid rulebook where the condition is violated, an exception should be raised in the rulebook initializer rather than here.

src/verifai/samplers/dynamic_rulebook_ce.py
dfremont: OK, then you should raise an exception when they create the rulebook, not here.

src/verifai/samplers/dynamic_rulebook_ce.py
dfremont: I don't think you changed this part of the code. Can you please explain why this is the intended logic? It looks wrong to me, and is inconsistent with how the old multi-objective CE sampler worked.

src/verifai/samplers/dynamic_rulebook_ce.py
dfremont: You didn't answer my question -- is this even necessary? If it is, catching any `Exception` is not the right way to do it.

src/verifai/samplers/dynamic_rulebook_ce.py
```diff
@@ -0,0 +1,141 @@
...
+            domain=domain,
+            buckets=self.cont_buckets,
+            dist=self.cont_dist,
+            alpha=self.alpha,
+            thres=self.thres,
+            priority_graph=priority_graph,
+        )
+        self.split_samplers = {}
+        for id, priority_graph in rb.priority_graphs.items():
+            self.split_samplers[id] = self.cont_ce(domain, priority_graph)
```
dfremont: These aren't SplitSamplers, they're just your ContinuousDynamicCESamplers. So the name is misleading, but more importantly, it looks like you're not handling the discrete part of the space at all. If that's an intentional limitation, we'll need to document it somewhere.

src/verifai/samplers/dynamic_rulebook_ce.py
```diff
@@ -0,0 +1,141 @@
...
+
+    def updateVector(self, vector, info, rho):
+        assert rho is not None
+        self.update_dist_from_multi(vector, info, rho)
+
+    def update_dist_from_multi(self, sample, info, rho):
+        try:
+            iter(rho)
+        except:
+            return
```
dfremont: Shouldn't use a bare `except`, and also returning silently here is surely not correct.

src/verifai/samplers/dynamic_rulebook_emab.py
```diff
@@ -0,0 +1,299 @@
...
+            priority_graph=priority_graph,
+            rulebook_instance=rb,
+        )
+        self.split_samplers = {}
+        for id, priority_graph in rb.priority_graphs.items():
+            self.split_samplers[id] = self.cont_ce(domain, priority_graph)
+        if not sorted(list(self.split_samplers.keys())) == list(
+            range(len(rb.priority_graphs))
+        ):
+            raise ValueError("Priority graph IDs should be in order and start from 0")
```
dfremont: See corresponding comments for the CE sampler.

src/verifai/samplers/dynamic_rulebook_emab.py
```diff
@@ -0,0 +1,299 @@
...
+                    to_remove.add(other_ce)
+            for other_ce in to_remove:
+                del self.counterexamples[other_ce]
+        self.counterexamples[ce] = np.array([np.zeros(int(b)) for b in self.buckets])
+        return True
+
+    def update_dist_from_multi(self, sample, info, rho):
+        try:
+            iter(rho)
+        except:
```
dfremont: Shouldn't be a bare `except`, and again, if this sampler is only for multi-objective falsification how can `rho` not be iterable?

src/verifai/samplers/dynamic_rulebook_emab.py
```diff
@@ -0,0 +1,299 @@
...
+                if self._compute_error_value(ce) > 0:
+                    print(
+                        "counterexamples =",
+                        ce,
+                        ", times =",
+                        int(
+                            np.sum(self.counterexamples[ce], axis=1)[0]
+                            / self._compute_error_value(ce)
+                        ),
+                    )
```
dfremont: Please use an f-string and move the big ratio onto its own line: then this whole thing should only take 2 lines and be a lot easier to read.

src/verifai/samplers/dynamic_rulebook_emab.py
```diff
@@ -0,0 +1,299 @@
...
+                proportions[0],
+                "\nsecond_term[0] =",
+                np.sqrt(self.exploration_ratio / self.counts * np.log(self.t))[0],
+                "\nratio[0] =",
+                proportions[0]
+                / (
+                    proportions
+                    + np.sqrt(self.exploration_ratio / self.counts * np.log(self.t))
+                )[0],
+            )
```
dfremont: Please split into one `print` call per line, using f-strings for each one.

src/verifai/samplers/dynamic_rulebook_emab.py
```diff
@@ -0,0 +1,299 @@
...
+            ranking_map[key] = count
+            count += value
+
+        self.error_weight = {}  # node_id -> weight
+        self.sum_error_weight = 0
+        for node in level:
+            self.error_weight[node] = ranking_map[level[node]]
+            self.sum_error_weight += 2 ** self.error_weight[node]
+        for key, value in sorted(self.error_weight.items()):
+            if self.verbosity >= 2:
```
dfremont:
```suggestion
        if self.verbosity >= 2:
            for key, value in sorted(self.error_weight.items()):
```

src/verifai/samplers/dynamic_rulebook_mab.py
```diff
@@ -0,0 +1,277 @@
...
+            thres=self.thres,
+            priority_graph=priority_graph,
+        )
+        self.split_samplers = {}
+        for id, priority_graph in rb.priority_graphs.items():
+            self.split_samplers[id] = self.cont_ce(domain, priority_graph)
+        if not sorted(list(self.split_samplers.keys())) == list(
+            range(len(rb.priority_graphs))
+        ):
+            raise ValueError("Priority graph IDs should be in order and start from 0")
```
dfremont: See corresponding comments for the other samplers.

src/verifai/samplers/dynamic_rulebook_mab.py
```diff
@@ -0,0 +1,277 @@
...
+                    to_remove.add(other_ce)
+            for other_ce in to_remove:
+                del self.counterexamples[other_ce]
+        self.counterexamples[ce] = np.array([np.zeros(int(b)) for b in self.buckets])
+        return True
+
+    def update_dist_from_multi(self, sample, info, rho):
+        try:
+            iter(rho)
+        except:
```
dfremont: See comment for EMAB sampler.

src/verifai/samplers/dynamic_rulebook_mab.py
dfremont, line 1: A lot of this code seems identical to the EMAB sampler. You don't have to do it now if it would take a while, but later you should make a common superclass for them to use.

src/verifai/samplers/multi_armed_bandit.py
```diff
@@ -170,19 +173,25 @@ def update_dist_from_multi(self, sample, info, rho):
...
-        # print(f'counter_ex = {counter_ex}')
-        # print(self.counterexamples)
         is_ce = self._add_to_running(counter_ex)
         for i, b in enumerate(info):
             self.counts[i][b] += 1.
             if is_ce:
                 self.counterexamples[counter_ex][i][b] += 1.
-        self.errors = self.invalid + self._get_total_counterexamples()
+        #self.errors = self.invalid + self._get_total_counterexamples()
+        self.errors = self._get_total_counterexamples()
```
dfremont: Was this a temporary change for debugging? I'm not sure what Kesav was trying to do with the `invalid` list (it seems like it shouldn't be possible to reach) -- if it's not needed any more, we should remove it entirely.

src/verifai/falsifier.py
```diff
@@ -176,15 +180,29 @@ def run_falsifier(self):
...
-                if self.save_error_table:
-                    self.populate_error_table(sample, rho)
-                ce_num = ce_num + 1
-                if ce_num >= self.ce_num_max:
-                    break
-            elif self.save_safe_table:
-                self.populate_error_table(sample, rho, error=False)
+            ce = False
+            if self.dynamic:
+                if isinstance(rho, int) and rho == 1:
```
dfremont, line 185: Why the special case for `1`? Are other non-list values possible? If so, they would crash in the `else` branch, right? If we're using a dynamic rulebook I think it's better to just have `rho` always be a list of lists.

src/verifai/falsifier.py
```diff
@@ -176,15 +180,29 @@ def run_falsifier(self):
...
+                    num_rule_outputs = sum(
+                        len(graph.nodes) for graph in self.monitor.priority_graphs.values()
+                    )
+                    concatenated_rho = [1.0] * num_rule_outputs
+                else:
+                    concatenated_rho = [value for rho_list in rho for value in rho_list]
+                self.populate_error_table(sample, concatenated_rho)
```
dfremont, line 192: You don't seem to be taking into account the `self.ce_num_max`, `self.save_error_table`, and `self.save_safe_table` options here (as the non-dynamic branch does).

tests/test_dynamic_rulebook_ce.py
```diff
@@ -0,0 +1,93 @@
+import networkx as nx
+import pytest
+from dotmap import DotMap
+
+from verifai.features import Box
+from verifai.samplers.dynamic_rulebook_ce import DynamicRulebookCrossEntropySampler
+
+
+class RulebookStub:
```
dfremont, line 9: Why not use actual Rulebooks in these tests, as a real user would?

tests/test_dynamic_rulebook_ce.py
dfremont, line 1: For each new type of sampler, there should be a test that creates a `FeatureSpace`, uses the appropriate `FeatureSampler` helper function to create the sampler, and generates some samples. See the existing tests for the ordinary CE and BO samplers for example.

tests/test_rulebook.py
```diff
@@ -0,0 +1,178 @@
...
+
+    def __init__(self, values):
+        self.values = np.array(values, dtype=float)
+        self.result = DummySimulation.Result(list(values))
+
+
+def test_rulebook_parses_rules_and_evaluates_segment(tmp_path):
+    rule_file = tmp_path / "toy_spec.py"
+    rule_file.write_text(
+        "\n".join(
```
dfremont, line 20: In all places like this, please use a triple-quoted string literal instead, and `inspect.cleandoc` if necessary to fix the indentation.

src/verifai/rulebook.py
```diff
@@ -0,0 +1,217 @@
...
+        exploration_ratio=2.0,
+        verbosity=1,
+        using_continuous=False,
+    ):
+        self.priority_graphs = {}
+        self.using_sampler = using_sampler
+        self.exploration_ratio = exploration_ratio
+        self.verbosity = verbosity
+        self.using_continuous = using_continuous
+        print("(rulebook.py) Parsing rules...")
```
dfremont: These and similar prints should be disabled when the verbosity is zero (which I would make the default, since people shouldn't have to set it manually in order not to get debugging messages).

src/verifai/rulebook.py
```diff
@@ -0,0 +1,217 @@
...
+        self.save_path = save_path
+
+    def _parse_rules(self, file_path):
+        with open(file_path, "r") as file:
+            file_contents = file.read()
+
+        tree = ast.parse(file_contents)
+        module_code = compile(tree, file_path, "exec")
+
+        namespace = {"__builtins__": __builtins__}
```
dfremont:
```suggestion
        namespace = {}
```
(there's no need to specify the builtins if you aren't customizing them)

src/verifai/rulebook.py
```diff
@@ -0,0 +1,217 @@
...
+                if not callable(function):
+                    raise ValueError(
+                        f"Rule function {node.name} is not callable in {file_path}"
+                    )
+                self.functions[node.name] = function
+
+        print(f"(rulebook.py) Parsed functions: {self.functions}")
+
+    def _parse_rulebooks(self, dir):
+        if os.path.isdir(dir):
```
dfremont, line 72: Surely if it's not a directory we don't want to just do nothing?

src/verifai/rulebook.py
```diff
@@ -0,0 +1,217 @@
...
+        if len(function_visitor.functions) == 0:
+            raise ValueError("No function found in segment function file")
+        if len(function_visitor.functions) > 1:
+            raise ValueError("Multiple functions found in segment function file")
+
+        function_node = function_visitor.functions[0]
+        function_code = compile(
+            ast.Module(body=[function_node], type_ignores=[]), "<string>", "exec"
+        )
+        exec(function_code)
```
dfremont: We don't want these functions executed in this local scope, right? I would pass in a namespace like you do above and extract the functions from it afterward.

src/verifai/error_table.py
dfremont: You didn't answer my question.

dfremont (changes requested):
There's still a large amount of duplicated code in the `collect_result` scripts and in the new samplers, and I think the work needed to set up the rulebooks could still be reduced, but we can work on that later. If you can work on the easier comments I've left above (including some I reopened from before since they haven't been addressed), I'll test this again and merge it.

examples/dynamic_rulebook/multi_02/multi_02.scenic
kevinchang73: Sure, I forgot to delete this one

examples/dynamic_rulebook/multi_03/util/multi_03_collect_result.py
kevinchang73: Yes, it's supposed to be 0,1,3,2

examples/dynamic_rulebook/multi_03/util/multi_03_collect_result.py
kevinchang73: I've simplified it.

kevinchang73: I also refactor other multi_*_collect_result.py

src/verifai/samplers/dynamic_rulebook_ce.py
kevinchang73: Yes, it's a limitation. It's mentioned in the README, and I'll also include it when we make the new documentation page.

src/verifai/samplers/dynamic_rulebook_ce.py
kevinchang73: Make sense. I should move this to where we parse the rulebook.

src/verifai/samplers/dynamic_rulebook_emab.py
kevinchang73: Sure. I guess it's formatted by the black package.

src/verifai/falsifier.py
kevinchang73: This is designed to handle the default Scenic rejectionFeedback (which is the integer 1)

src/verifai/falsifier.py
kevinchang73: No, I don't. For the dynamic rulebook, the number of iterations is controlled by `multi.py`, and all the samples are written to a single table.

src/verifai/rulebook.py
kevinchang73: I've added an `if verbosity >= 1:` block for each print statement. However, I'd prefer to keep the default verbosity at 1 so the evaluation results are still printed, since they are needed for the result collection scripts (e.g., `multi_inter_left_collect_result.py`).

src/verifai/rulebook.py
kevinchang73: I add 
`else:
            raise ValueError(
                f"Rulebook graph path must be an existing directory."
            )`

src/verifai/samplers/dynamic_rulebook_ce.py
kevinchang73: It's a heuristic to adapt the CE sampler to rulebook specifications. If we define a counterexample as a sample that violates _any_ rule, the CE sampler might be guided to only violate low-priority rules. Thus, I designed it this way to force the sampler to target high-priority rule violations instead.

src/verifai/samplers/multi_armed_bandit.py
kevinchang73: It seems that the `invalid` list is also used for the default Scenic `rejectionFeedback`. However, I still believe `self.errors` should only consider `self._get_total_counterexamples()`.

src/verifai/error_table.py
kevinchang73: I think it's fine to remove the change.

src/verifai/samplers/dynamic_rulebook_emab.py
kevinchang73: Switch to `except TypeError:` for now.

src/verifai/falsifier.py
dfremont: OK, we should probably change the rejection feedback in the multi-objective case, but we can discuss that later.

src/verifai/falsifier.py
dfremont: The documented interface of the falsifier class should not change just because you're using a dynamic rulebook instead of some other kind of monitor. Also `multi.py` is just an example file and you can't assume everybody using rulebooks will use it. But we can merge this as-is and you can add proper handling of these options before the official release.

dfremont (approved):
Thanks again, Kai-Chun. This is looking OK to merge now, but let's discuss the outstanding comments on Tuesday.

PR #60, opened 2026-03-04
Initial version of ODD monitoring code and examples

a-luque:


examples/modd/carla/models/controller_cte_dist_130.pth
dfremont, line 1: It's not ideal to add a 50 MB data file to the repository -- is there any way we can make this smaller? If not (e.g. it's not feasible to use a smaller model, or having users train their own would take a long time / a fancy GPU), I guess it's OK.

examples/modd/carla/controller_training.py
```diff
@@ -0,0 +1,225 @@
+import pandas as pd
+import numpy as np
+import os
+import torch
+import torchvision
```
dfremont, line 5: Dependencies like these which are not already listed in `pyproject.toml` should be added there under the `examples` extra (so that someone running `pip install verifai[examples]` gets everything necessary to run the examples -- please test this in a fresh virtual environment).

examples/modd/carla/controller_training.py
```diff
@@ -0,0 +1,225 @@
...
+        if test_loss < best_val_loss:
+            best_val_loss = test_loss
+            torch.save(model.state_dict(), os.path.join(save_path, f"{model_name}_{epoch}.pth"))
+        if (epoch % 50 == 0 or epoch == n_epochs-1) and epoch > 50 :
+            torch.save(model.state_dict(), os.path.join(save_path, f"{model_name}_{epoch}.pth"))
+    return model
+
+
+from torch.utils.data import Dataset
+from torchvision.io import read_image
```
dfremont, line 122: Please move to the top of the file with the rest of the imports so that they are more easily found (likewise for the other imports later on).

examples/modd/carla/controller_training.py
```diff
@@ -0,0 +1,225 @@
...
+parser.add_argument("--data_dir", type=str, default="./training_data/")
+parser.add_argument("--model_dir", type=str, default="./models")
+parser.add_argument("--model_name", type=str, default="controller_cte_dist")
+args = parser.parse_args()
+
+if __name__ == "__main__":
+
+    import os
+
+    sys.setrecursionlimit(10000)
```
dfremont, line 212: Is this really necessary? It's not a good sign to have to change the recursion limit, and nothing here looks recursive on a quick skim.

examples/modd/carla/controller_training.py
```diff
@@ -0,0 +1,225 @@
+import pandas as pd
```
dfremont, line 1: In all files, please separate imports into 3 sections separated by 1 blank line: Python stdlib imports, external packages (like `pandas`), and local imports from within VerifAI.

examples/modd/carla/data_generation.py
```diff
@@ -0,0 +1,76 @@
...
+import argparse
+import pickle
+import numpy as np
+import time
+import scenic
+from dotmap import DotMap
+from PIL import Image
+import cv2
+
+import pandas as pd
```
dfremont, line 16: These and some of the other imports don't appear to be used.

examples/modd/carla/data_generation.py
```diff
@@ -0,0 +1,76 @@
...
+        if os.path.exists(data_dir):
+            shutil.rmtree(data_dir)
+        os.mkdir(data_dir)
+
+        
+    sim_dir = f"{data_dir}{i}"
+    if not os.path.exists(sim_dir):
+        os.mkdir(sim_dir)
+
+    flag_good = False
```
dfremont, line 61: This is never used.

examples/modd/carla/data_generation.py
```diff
@@ -0,0 +1,76 @@
...
+    print(f"Executing sim {i}")
+    try:
+        os.system(f"scenic -S {scenic_path} --count 1 --time {num_of_steps} --2d --param recordFolder {sim_dir} --param timeout 30")
```
dfremont, line 64: Using Scenic's Python API would probably be better here, but if you want to stick with the CLI, you should use `subprocess.run` with a list of arguments to make sure they are properly quoted/escaped on all platforms.

examples/modd/carla/data_generation.py
```diff
@@ -0,0 +1,76 @@
...
+        flag_good = True
+    except:
```
dfremont, line 66: Never use a bare `except:`, which will catch `KeyboardInterrupt` and prevent people from stopping your script with Ctrl-C. You should use `except Exception:` if you have no more specific exception types in mind.

examples/modd/carla/data_generation.py
```diff
@@ -0,0 +1,76 @@
...
+        os.mkdir(sim_dir)
+
+    flag_good = False
+    print(f"Executing sim {i}")
+    try:
+        os.system(f"scenic -S {scenic_path} --count 1 --time {num_of_steps} --2d --param recordFolder {sim_dir} --param timeout 30")
+        flag_good = True
+    except: 
+        print(f"Simulation {j} failed.")
+        wait(5)
```
dfremont, line 68: I don't think `wait` is defined.

examples/modd/carla/data_generation.py
```diff
@@ -0,0 +1,76 @@
...
+        os.system(f"scenic -S {scenic_path} --count 1 --time {num_of_steps} --2d --param recordFolder {sim_dir} --param timeout 30")
```
dfremont, line 69: If the first attempt failed, why would you expect it to work a second time after a 5-second delay? Scenic already retries the simulation if there was a `SimulationCreationError`, and any other ways Scenic might fail (e.g. if there is a syntax error in the program) are unlikely to fix themselves given a little time. Given that `wait` isn't defined, I suspect this code path was never run: in which case you might as well remove it.

examples/modd/carla/data_generation.py
```diff
@@ -0,0 +1,76 @@
...
+        
+
+
+        
+
+
+
```
dfremont, line 76: Please remove unnecessary blank lines (1 is OK).

examples/modd/carla/followLeader_datagen.scenic
```diff
@@ -0,0 +1,245 @@
+""" Scenario Description
+The ego car follows the leader car maintaining a normal distance while lane keeping.
+"""
+import pickle
+import sklearn
+import numpy as np 
+import random
```
dfremont, line 7: Scenic programs should not use the `random` module directly: use Scenic's own distribution facilities for generating random numbers (otherwise you will break serialization, for example).

examples/modd/carla/followLeader_datagen.scenic
```diff
@@ -0,0 +1,245 @@
...
+EGO_MODEL = "vehicle.tesla.model3"
+LEADER_SPEED = Range(8,12)
+EGO_SPEED = 5
+THROTTLE_ACTION = 0.6
+BRAKE_ACTION = 1.0
+EGO_TO_LEADER = Range(-15, -10)
+EGO_BRAKING_THRESHOLD = 10
+
+
+behavior FollowLaneBehaviorModified(target_speed = 10, laneToFollow=None, is_oppositeTraffic=False, leaderCar=None):
```
dfremont, line 33: It's unfortunate to have to duplicate most of the code of `FollowLaneBehavior` here. It's probably unavoidable, since it looks like you're adding noise to the actions the ordinary behavior would take, but in that case I would just put a short docstring explaining how this version differs from the ordinary one rather than copying the original docstring).

examples/modd/carla/followLeader_datagen.scenic
```diff
@@ -0,0 +1,245 @@
...
+attrs = {"image_size_x": 640,
+         "image_size_y": 320}
+
+start = new OrientedPoint on lane.centerline
+leader = new Car at start,
+            with blueprint EGO_MODEL,
+            with behavior FollowLaneBehaviorModified(target_speed=EGO_SPEED),
+            with color Color(0,0,0) 
+
+ego = new Car following roadDirection from leader.position for EGO_TO_LEADER,
```
dfremont, line 216:
```suggestion
ego = new Car following roadDirection from leader for EGO_TO_LEADER,
```

examples/modd/carla/followLeader_datagen.scenic
```diff
@@ -0,0 +1,245 @@
...
+            with blueprint EGO_MODEL,
+            with behavior FollowLaneBehaviorModified(target_speed=EGO_SPEED, leaderCar=leader),
+            with cte 0,
+            with sensors {"front_rgb": RGBSensor(offset=(0, 2, 1), attributes=attrs)
+                        }
```
dfremont, line 221:
```suggestion
            with sensors {"front_rgb": RGBSensor(offset=(0, 2, 1), attributes=attrs)} 
```

examples/modd/carla/followLeader_datagen.scenic
```diff
@@ -0,0 +1,245 @@
...
+            with blueprint EGO_MODEL,
+            with behavior FollowLaneBehaviorModified(target_speed=EGO_SPEED, leaderCar=leader),
+            with cte 0,
+            with sensors {"front_rgb": RGBSensor(offset=(0, 2, 1), attributes=attrs)
+                        } 
+
+
+
+
+require distance to intersection > 10 and distance from leader to intersection > 10 and ego can see leader
```
dfremont, line 226: Better to write this as 3 separate `require` statements: it will help Scenic's pruning, and it's more obvious that you have 3 independent constraints.

Actually, for `ego can see leader` it would be better to just add `with requireVisible True` when defining `leader` (which will also help Scenic do pruning).

examples/modd/carla/followLeader_extracar.scenic
```diff
@@ -0,0 +1,491 @@
+""" Scenario Description
+The ego car follows the leader car maintaining a normal distance while lane keeping.
```
dfremont, line 2: This is the same description as the previous file.

examples/modd/carla/followLeader_extracar.scenic
```diff
@@ -0,0 +1,491 @@
...
+"""
+
+import warnings
+warnings.filterwarnings("ignore")
```
dfremont, line 6: Why? If actually necessary, can you use a more specific filter?

examples/modd/carla/followLeader_extracar.scenic
```diff
@@ -0,0 +1,491 @@
...
+import sys
+import cv2
+from scenic.domains.driving.controllers import (
+    PIDLateralController,
+    PIDLongitudinalController,
+)
+
+if globalParameters.seed != "":
+    random.seed(globalParameters.seed)
+    np.random.seed(globalParameters.seed)
```
dfremont, line 28: Should not need this: use Scenic's `-s` / `--seed` option.

examples/modd/carla/followLeader_extracar.scenic
```diff
@@ -0,0 +1,491 @@
...
+
+torch.set_default_device('cuda')
+
+class resNet(torch.nn.Module):
```
dfremont, line 32: Aren't this and the following class the same as in `controller_training.py`? Please consolidate if possible.

examples/modd/carla/followLeader_extracar.scenic
```diff
@@ -0,0 +1,491 @@
...
+
+monitor_model = None
+if globalParameters.monitor != "":
+    with open(globalParameters.monitor, 'rb') as f:
+        monitor_model = pickle.load(f) 
+   
+    
+
+
+behavior FollowLaneBehaviorModified(target_speed = 10, laneToFollow=None, is_oppositeTraffic=False, leaderCar=None):
```
dfremont, line 91: If this is basically the same as in the other Scenic file but with the invocation of the CNN added, maybe you can define the behavior once in a helper module with an option to pass in the CNN and then import it in the scenarios that need it?

dfremont: Also same comment as before about having the docstring reflect what this actually does (instead of the original `FollowLaneBehavior`).

examples/modd/carla/followLeader_extracar.scenic
```diff
@@ -0,0 +1,491 @@
...
+        if leaderCar is None:
+            self.steps_speed += 1
+            if self.steps_speed > 20:
+                original_target_speed = LEADER_SPEED.getSample()
+                self.steps_speed = 0
+            
+        else:
+            steps_running += 1
+
+        if leaderCar and monitor_model and steps_running > 80:
```
dfremont, line 295: The code here looks the same as the part starting on line 240 -- can you consolidate into a helper function?

examples/modd/carla/followLeader_extracar.scenic
```diff
@@ -0,0 +1,491 @@
...
+            print("-- Monitor: Not safe! Switching to SafeBehavior.")
+            take SetBrakeAction(1.0)
+            do EgoBehavior2(target_speed = target_speed, controller_path = controller_path, leaderCar=leaderCar, obstacleCar=obstacleCar)
+
+behavior EgoBehavior2(target_speed = 10, controller_path = None, leaderCar=None, obstacleCar=None):
+    try:
+        do FollowLaneBehaviorModified(target_speed=target_speed, leaderCar=leader)
+    interrupt when self.isSafe:
+        print("-- Monitor: Safe! Switching back to CNN Controller.")
+        do EgoBehavior(target_speed=target_speed, controller_path=controller_path, leaderCar=leaderCar, obstacleCar=obstacleCar)
```
dfremont, line 325: This approach of calling each version of the behavior from the other will lead to the stack getting deeper and deeper each time you switch (and Scenic will need to monitor more and more interrupt conditions). Why not just `return` instead of `do EgoBehavior(...)`? You could even drop the interrupt statement and just write `do FollowLaneBehaviorModified(...) until self.isSafe`, at which point this behavior becomes so short you might as well inline it into the previous one.

examples/modd/carla/followLeader_extracar.scenic
```diff
@@ -0,0 +1,491 @@
...
+    controller.cuda()
+    controller.load_state_dict(torch.load(controller_path))
+    controller.eval()
+
+    # Features for monitor
+    if monitor_model:
+        color = obstacle.color
+        self.r,self.g,self.b = color.r, color.g, color.b
+        self.weather = np.array([
+            self.carlaActor.get_world().get_weather().cloudiness,
```
dfremont, line 358: Can you just call `self.carlaActor.get_world().get_weather()` once to reduce all this boilerplate?

examples/modd/carla/followLeader_extracar.scenic
```diff
@@ -0,0 +1,491 @@
...
+            self.carlaActor.get_world().get_weather().wetness,
+            self.carlaActor.get_world().get_weather().fog_falloff,
+            self.carlaActor.get_world().get_weather().scattering_intensity,
+            self.carlaActor.get_world().get_weather().mie_scattering_scale,
+            self.carlaActor.get_world().get_weather().rayleigh_scattering_scale,
+            self.carlaActor.get_world().get_weather().dust_storm  
+        ])
+
+    while True:
+        front_img = self.sensors["front_rgb"]._lastObservation
```
dfremont, line 375: Any reason to use this internal, undocumented API? Why not `self.observations["front_rgb"]`?

examples/modd/carla/followLeader_extracar.scenic
```diff
@@ -0,0 +1,491 @@
...
+        if dist_pred < EGO_BRAKING_THRESHOLD:
+            take SetBrakeAction(1.0)
+        else:
+            take RegulatedControlAction(throttle, current_steer_angle, past_steer_angle)
+        past_steer_angle = current_steer_angle
+        past_speed = current_speed
+        
+
+        # Monitor trigger
+        if monitor_model:
```
dfremont, line 413: The body of this conditional can probably be moved into a helper function used in 3 places as suggested above.

examples/modd/carla/followLeader_extracar.scenic
```diff
@@ -0,0 +1,491 @@
...
+            with blueprint EGO_MODEL,
+            with behavior FollowLaneBehaviorModified(target_speed=LEADER_SPEED),
+            with steps_speed 0,
+            with color Color(0,0,0),
+            with time_steps 0
+
+obstacle = new Car at 0 @ 25 relative to leader,
+            with heading leader 
+
+ego = new Car following roadDirection from leader.position for EGO_TO_LEADER,
```
dfremont, line 452:
```suggestion
ego = new Car following roadDirection from leader for EGO_TO_LEADER,
```

examples/modd/carla/followLeader_extracar.scenic
```diff
@@ -0,0 +1,491 @@
...
+            with blueprint EGO_MODEL,
+            with behavior EgoBehavior(target_speed=EGO_SPEED, controller_path=globalParameters.controller, leaderCar=leader, obstacleCar=obstacle),
+            with cte 0,
+            with distIntersec EGO_TO_LEADER,
+            with isSafe 1,
+            with sensors {"front_rgb": RGBSensor(offset=(0, 2, 1), width=640, height=320),
+                        "aerial_rgb": RGBSensor(offset=(0, -10, 4), width=1280, height=640)
```
dfremont, line 459: This sensor doesn't appear to be used.

examples/modd/carla/followLeader_extracar.scenic
```diff
@@ -0,0 +1,491 @@
...
+                        },
+
+
+
+require ego can see leader
```
dfremont, line 464: See note about `requireVisible` in previous file.

examples/modd/carla/followLeader_extracar_NN.scenic
```diff
@@ -0,0 +1,516 @@
+""" Scenario Description
+The ego car follows the leader car maintaining a normal distance while lane keeping.
```
dfremont, line 2: Again, please distinguish from other scenarios.

examples/modd/carla/followLeader_extracar_NN.scenic
```diff
@@ -0,0 +1,516 @@
...
+"""
+
+import warnings
+warnings.filterwarnings("ignore")
```
dfremont, line 6: I won't keep repeating the comments -- please check all from the previous file in this one too.

examples/modd/carla/followLeader_extracar_NN.scenic
```diff
@@ -0,0 +1,516 @@
...
+    monitor_model = MLP()
+    monitor_model.cuda()
+    monitor_model.load_state_dict(torch.load(globalParameters.monitor))
+    monitor_model.eval()
+    print("Monitor loaded")
+   
+    
+
+
+behavior FollowLaneBehaviorModified(target_speed = 10, laneToFollow=None, is_oppositeTraffic=False, leaderCar=None):
```
dfremont, line 112: This version looks almost identical to the one in the previous file -- please consolidate to make these scenarios easier to follow.

examples/modd/carla/static_sensors.py
```diff
@@ -0,0 +1,51 @@
+import os
+from time import sleep
+
+import carla
+import cv2
+import numpy as np
+
+from scenic.core.sensors import ActiveSensor
+
+class CarlaOtherFeaturesSensor(ActiveSensor):
```
dfremont, line 10: I don't think this is actually used anywhere. Maybe this file is left over from some other experiments?

examples/modd/modd_learner_main.py
```diff
@@ -0,0 +1,215 @@
...
+
+eval_params.save_model_path = trainer_params.save_model_path
+
+
+
+#####################################################################
+#################### Sampling parameters ############################
+#####################################################################
+
+class SpecMonitor(specification_monitor):
```
dfremont, line 145: Just as a heads-up, some minor changes will be needed to this code when the simplification of monitors is merged -- but I just realized while reviewing this that I forgot to make a PR for that last December. Once I've made that PR I can try fixing this up myself.

src/verifai/modd/odd_data_generator.py
```diff
@@ -0,0 +1,128 @@
...
+        if self.datagen_params.verbosity >= 2:
+            print(f'Server class is {type(self.sampling_params.server)}')
+
+        if self.datagen_params.verbosity >= 1:
+            bar = progressbar.ProgressBar(max_value=num_simulations)
+
+        try:
+            while True:
+                try:
+                    start = time.time()
```
dfremont, line 53: Should use `time.monotonic()` for all timings so that they aren't messed up by changes to the system clock.

src/verifai/modd/odd_data_generator.py
```diff
@@ -0,0 +1,128 @@
...
+                            else self.sampling_params.server.monitor.evaluate(result))
+                    self.sampling_params.server.lastValue = value
+                    
+                    after_simulation = time.time()
+                    timings = ServerTimings(sample_time=(after_sampling - start),
+                                            simulate_time=(after_simulation - after_sampling))
+                    rho = self.sampling_params.server.lastValue
+                    self.total_sample_time += timings.sample_time
+                    self.total_simulate_time += timings.simulate_time
+                except:
```
dfremont, line 73: Don't use bare `except`, and please distinguish the error and non-error paths (for the latter I assume you want `TerminationException`?).

src/verifai/modd/odd_data_generator.py
```diff
@@ -0,0 +1,128 @@
...
+                    scene = self.sampling_params.server.sampler.lastScene
+                    assert scene
+                    result = self.sampling_params.server._simulate(scene)
+                    if result is None:
+                        print(self.sampling_params.server.rejectionFeedback)
+                        return self.sampling_params.server.rejectionFeedback
+                    print(f"Time steps run for this simulation: {len(result.trajectory)}")
+                    value = (0 if self.sampling_params.server.monitor is None
+                            else self.sampling_params.server.monitor.evaluate(result))
+                    self.sampling_params.server.lastValue = value
```
dfremont, line 65: This block seems to replicate the logic of `Server.run_server` (and `ScenicServer.evaluate_sample`), including manipulating internal state of the server, which is fragile. Can't you just call `run_server` on the `Server` you have?

src/verifai/modd/odd_data_generator.py
```diff
@@ -0,0 +1,128 @@
...
+            if self.datagen_params.verbosity >= 1:
+                bar.finish()
+            self.sampling_params.server.terminate()
+        if self.datagen_params.verbosity >= 1:
+            print('All simulations generated.')
+        print(f"Total of {len(self.samples)} simulations generated.")
+        return 0
+    
+    def load_simulations(self, load_path):
+        import os
```
dfremont, line 106: Unnecessary, and if it was needed it should be at the top of the file.

src/verifai/modd/odd_data_generator.py
```diff
@@ -0,0 +1,128 @@
...
+        self.training_data = None
+        self.samples = {}
+
+    def sample_simulations(self):
+        pass
+
+    def generate(self):
+        pass
+
+class GenericDataGenerator(DataGenerator):
```
dfremont, line 24: There don't seem to be any other subclasses of `DataGenerator`, so why not just move all of this code into `DataGenerator` and drop this class?

src/verifai/modd/odd_evaluator.py
```diff
@@ -0,0 +1,135 @@
...
+        self.terminate = False
+        self.evals = []
+        self.samples = {}
+
+    def evaluate(self, monitor):
+        pass
+
+
+
+class GenericEvaluator(Evaluator):
```
dfremont, line 25: What's the need for this class vs. just moving everything into `Evaluator`?

src/verifai/modd/odd_evaluator.py
```diff
@@ -0,0 +1,135 @@
...
+    def __init__(self, eval_params, sampling_params, global_params):
+        super().__init__(eval_params, sampling_params, global_params)
+
+
+    def sample_simulations(self, num_simulations, num_steps, save_datagen_path, save_scenes_path):
```
dfremont, line 30: This seems very similar to `DataGenerator.sample_simulations`. Please consolidate the common code.

src/verifai/modd/odd_data_generator.py
```diff
@@ -0,0 +1,128 @@
...
+                if self.datagen_params.verbosity >= 2:
+                    print("Sample no: ", i, "\nSample: ", sample, "\nRho: ", rho, "\nRecords: ", result.records)
+                self.samples[i] = (result.records, rho)
+                if self.datagen_params.verbosity >= 1:
+                    bar.update(i)
+                i += 1
+                if i == 1:
+                    t0 = time.time()
+                if i == num_simulations:
+                    filehandler = open(f"{save_path}training_{i}.pkl", 'wb')
```
dfremont, line 88: It doesn't look like this file will ever be closed, nor is it used anywhere else. I would use a `with` statement. (Ditto for the similar code a few lines down.)

src/verifai/modd/odd_data_generator.py
```diff
@@ -0,0 +1,128 @@
...
+    def load_simulations(self, load_path):
+        import os
+        with open(load_path, 'rb') as f:
+            return pickle.load(f)
+
+    def generate(self, num_simulations, num_steps):
+        self.init_sampler()
+        save_path=self.datagen_params.datagen_save_dir
+        if save_path[-1] != "/":
+            save_path += "/"
```
dfremont, line 114: This won't work on Windows. Use `pathlib` or `os.path` to construct paths instead.

src/verifai/modd/odd_evaluator.py
dfremont: Actually, the entire class is almost the same as the data generator, so I would make a common superclass of those (which does not need to be public).

src/verifai/modd/odd_learner.py
```diff
@@ -0,0 +1,45 @@
...
+        self.trainer_params = trainer_params
+        self.eval_params = eval_params
+        self.sampling_params = sampling_params
+        self.global_params = global_params      
+
+    def run(self):
+        pass
+
+
+class MODDLearner(ODDLearner):
```
dfremont, line 26: What is the difference supposed to be between these two classes? Unless you anticipate further subclasses, I would just have a single class.

src/verifai/modd/odd_monitor.py
```diff
@@ -0,0 +1,164 @@
...
+
+        # loop over epochs
+        for epoch in range(1, self.num_epochs + 1):
+            # training loop
+            train_loss = self.train_loop(epoch, dataloader,  optimizer, loss_function, self.device)
+
+            # collect losses
+            train_loss_all.append(train_loss)
+
+            # Print message
```
dfremont, line 116: Can we drop all these obvious, uninformative comments? (presumably AI generated?)

src/verifai/modd/odd_trainer.py
```diff
@@ -0,0 +1,39 @@
...
+    def __init__(self, trainer_params):
+        self.trainer_params = trainer_params
+        self.monitor = None
+        self.training_results = None
+
+    def train(self, training_data):
+        pass
+    
+
+class GenericTrainer(Trainer):
```
dfremont, line 20: I don't see the need for both of these classes.

src/verifai/modd/pymodd.py
```diff
@@ -0,0 +1,68 @@
...
+        return self.monitor, self.training_results, self.evaluation_results
+    
+
+
+    
+    
+
+    
+
+
```
dfremont, line 68: Please remove extraneous whitespace.

src/verifai/modd/pymodd.py
```diff
@@ -0,0 +1,68 @@
+from verifai.modd.odd_evaluator import GenericEvaluator
+from verifai.modd.odd_data_generator import GenericDataGenerator
+from verifai.modd.odd_trainer import GenericTrainer
+from verifai.modd.odd_monitor import Monitor
+from dotmap import DotMap
+
+class MODD():
```
dfremont, line 7: Should this actually be a class? You only `run` it once, I think, so if there's no reason to subclass it, it might as well just be a function. A more informative name like `generate_monitor` might be better too.

src/verifai/modd/odd_sampler.py
```diff
@@ -0,0 +1,37 @@
+from dotmap import DotMap
+from verifai.samplers import ScenicSampler
+import os
+
+class ODDSampler():
+    def __init__(self, sampling_params):
+        self.sampling_params = sampling_params
+
+        if self.sampling_params.has_key("server_type"):
```
dfremont, line 9: Can this condition ever be false?

src/verifai/modd/odd_sampler.py
```diff
@@ -0,0 +1,37 @@
...
+            if self.sampling_params.server_type == "scenic":
```
dfremont, line 10: If I understand the code correctly, it only works with the Scenic server anyway, so why do this check? Raising an exception if the user tries to instantiate `MODD` with a different kind of server would make more sense. In any case, silently doing nothing when this condition is false seems like the wrong thing to do.

src/verifai/modd/odd_sampler.py
```diff
@@ -0,0 +1,37 @@
...
+        if not self.sampling_params.has_key("sampler_type"):
+            self.sampler_type = 'random'
+        sampling_data.sampler_type = self.sampler_type
+        sampling_data.sample_space = None
+        sampling_data.sampler_params = None
+        if self.sampling_params.mode == "train":
+            sampling_data.sampler = ScenicSampler.fromScenario(self.sampling_params.path, mode2D=True, maxSteps=300, params={"monitor": "", "seed":"", "render" : 1, "verbosity": 3, "timeBound": 300, "controller": self.sampling_params.controller_path})
+            
+        elif self.sampling_params.mode == "eval":
+            sampling_data.sampler = ScenicSampler.fromScenario(self.sampling_params.path, mode2D=True, maxSteps=300, params={"monitor": os.path.abspath(self.sampling_params.monitor), "seed": 42, "render" : 1, "verbosity": 3, "timeBound": 300, "controller": self.sampling_params.controller_path})
```
dfremont, line 30: Do you really want to hard-code the seed and the number of steps/timeBound here?

Also, why force `mode2D`? I don't think anything you're doing requires that, and it's possible to run CARLA scenarios in 3D mode for example.

src/verifai/modd/odd_sampler.py
```diff
@@ -0,0 +1,37 @@
...
+            
+        elif self.sampling_params.mode == "eval_nomonitor":
+            sampling_data.sampler = ScenicSampler.fromScenario(self.sampling_params.path, mode2D=True, maxSteps=300, params={"monitor": "", "seed": 42, "render" : 1, "verbosity": 3, "timeBound": 300, "controller": self.sampling_params.controller_path})
```
dfremont, line 33: The differences between these three cases would be more evident if you had them modify a common set of parameters, then call `ScenicSampler.fromScenario` once after the `if` statement.

Also you should have either an assertion or an exception to handle the case of an invalid `mode` (which one depends on whether the user can specify the mode or whether it's a purely internal thing).

examples/modd/modd_learner_main.py
```diff
@@ -0,0 +1,215 @@
...
+import os.path
+import sys
+import numpy as np
+import torch.nn as nn
+from sklearn.tree import DecisionTreeClassifier
+from sklearn.neural_network import MLPClassifier
+from sklearn.pipeline import Pipeline
+from sklearn.linear_model import LogisticRegression
+from sklearn.preprocessing import StandardScaler
+from verifai.modd.odd_sampler import ODDSampler
```
dfremont, line 15: Not used.

src/verifai/modd/pymodd.py
dfremont: Actually I now realize that `MODDLearner` is the top-level API. But it doesn't seem to do anything itself, just call this class. Why do you need both?

dfremont (changes requested):
Hi Alejandro, thanks for putting this together! The code looks OK overall, but see the detailed comments above. A few bigger things to change/investigate:

* I could not run the example on my laptop because it requires CARLA. I can try later on another machine, but it would be nice to have a more portable example. I don't think anything you're doing is CARLA-specific, so could you try making it work with MetaDrive too? I spent a few minutes on that myself but ran into this error:
```
  File "/Users/dfremont/Documents/Berkeley/repositories/BerkeleyLearnVerify/VerifAI/src/verifai/modd/odd_data_generator.py", line 48, in sample_simulations
    bar = progressbar.ProgressBar(max_value=num_simulations)
TypeError: __init__() got an unexpected keyword argument 'max_value'
```
* Some unit tests for the new functionality should be added. Even a basic test (e.g. using the Newtonian simulator) would catch issues like forgetting to add a dependency to `pyproject.toml`. (For example, the code will currently not run because `tqdm` is missing.)
* I think the `MODDLearner` class is the only one that's intended to be public (?), but it's not clear how to actually use it. The README in the example explains what it does mathematically, but there's no documentation of the actual API (e.g. how does one actually run the resulting monitor?). A new page should be added to the documentation explaining the monitor generation feature.
* Even if all the other classes are internal, it would be good to at least have short docstrings so that people maintaining this code can understand what each of the components is supposed to do, what its arguments are, etc. Right now the internal architecture is not clear.

pyproject.toml
```diff
@@ -61,6 +61,9 @@ examples = [
 	"pillow >= 9.3",
 	'pyproj ~= 3.0; python_version < "3.10"',
 	'pyproj ~= 3.3; python_version >= "3.10"',
+	"torch ~= 2.10.0",
+	"torchvision ~= 0.25.0",
+	"carla ~= 0.9.16",
```
dfremont: We can't add `carla` here, since it isn't available on all platforms we support.

pyproject.toml
```diff
@@ -61,6 +61,9 @@ examples = [
 	"pillow >= 9.3",
 	'pyproj ~= 3.0; python_version < "3.10"',
 	'pyproj ~= 3.3; python_version >= "3.10"',
+	"torch ~= 2.10.0",
+	"torchvision ~= 0.25.0",
```
dfremont, line 66: Do you need specific minor versions here, or would torch 2.11 be OK for example? If the latter, you should put `torch ~= 2.10` so the minor version is not upper bounded.

examples/modd/carla/followLeader_extracar.scenic
```diff
@@ -1,312 +1,61 @@
...
 
+from followCarBehaviorMODD import (
+    run_MODD,
+    FollowCarBehaviorMODD,
+)
+
 #CONSTANTS
 EGO_MODEL = "vehicle.tesla.model3"
-LEADER_SPEED = TimeSeries(VerifaiRange(6,8))
+LEADER_SPEED = Range(6,8) # TimeSeries(VerifaiRange(6,8))
```
dfremont, line 38: Was this a temporary change for debugging? In the paper we state that we use a time series (sampled every 2 seconds), in order to demonstrate the new support for time series features.

examples/modd/carla/followLeader_extracar.scenic
dfremont: I think this concern is still valid.

src/verifai/modd/odd_data_generator.py
dfremont: I don't think this has been resolved.

src/verifai/modd/odd_data_generator.py
dfremont: What you have now is better, but I think you're still replicating `ScenicServer.evaluate_sample`.

src/verifai/modd/odd_sampler.py
```diff
@@ -0,0 +1,39 @@
...
+    def __init__(self, sampling_params):
+        self.sampling_params = sampling_params
+
+        try:
+            server_params = DotMap()
+            server_params.update(self.sampling_params.server_options)
+            self.init_server(server_params, self.sampling_params.server_class)
+            self.sampling_params.server = self.server
+        except Exception:
+            print("ODDSampler initialization failed. Is the server a Scenic server?")
```
dfremont, line 15: I don't think we want to suppress all exceptions here and silently end up with a not-properly-initialized `ODDSampler`. Why catch exceptions at all, instead of letting them bubble up? If there are constraints on the type of server that is allowed for `ODDSampler`, for example, it would be better to check for violations of those specifically and raise a specific error message.

dfremont (changes requested):
Thanks, Alejandro, this is looking much better! We can leave tests and further cleanup for later as we discussed, but there are a few things that still look suspect to me: please see the comments I left above, including a few from before that I've added a reply to. I will try to test the code again tomorrow.

dfremont:
@a-luque Now when I run the example it gets stuck in an infinite loop saying `Sampling failed.` On line 46 of `odd_data_generator.py` the expression `self.sampling_params.server` does not evaluate to an actual `Server`, so calling `run_server()` on it fails, and then since you catch all exceptions and retry we get stuck.

dfremont:
Actually it looks like the bug above is only triggered when the `carla` package isn't installed. If I try it with CARLA 0.9.14, it segfaults, but with CARLA 0.9.16 it works (although it prints a lot of warnings and an unfortunate `ODDSampler initialization failed. Is the server a Scenic server?` message).

dfremont:
I'm going to merge this so I can prepare a branch with everything in it, but let's come back to the unresolved comments above.

examples/modd/carla/followLeader_extracar.scenic
a-luque: Yes! Sorry for that. I didn't manage to get TimeSeries to work when creating a new environment with just VerifAI.

I did some tests, and I didn't manage to get it to work. I guess it is because I was using the VerifAIDynamicSampling branch of Scenic to get access to TimeSeries, and the Scenic version that VerifAI currently installs does not match it. I am not sure about how to explicitly ask for it to be installed.

dfremont: For the record, we resolved this in the `v3` branch.

PR #61, opened 2026-03-31
Use pygame-ce on Python 3.10+

lola831:
Scenic is moving to `pygame-ce` for Python 3.10+, since `pygame-ce` provides support for newer Python versions.
VerifAI needs the same conditional dependency change because Scenic and VerifAI depend on each other. This matches the corresponding Scenic change in [PR #371](https://github.com/BerkeleyLearnVerify/Scenic/pull/371).


dfremont (approved):
Thanks Lola! I will make a new patch release of both Scenic and VerifAI so people consistently get `pygame-ce` from now on.

PR #62, opened 2026-05-11
Preparation for v3

dfremont:


dfremont:
Per discussion with Eric, we will merge this into `v3` for now and review it before the release.

PR #65, opened 2026-07-01
Replace progressbar2 with tqdm

Eric-Vin:
Tested `falsifier.py` via `examples/scenic/falsify_distance.py`, but wasn't sure how to test the `ParallelScenicServer` which was also updated.


dfremont (approved):
Looks good, thanks Eric!

